PR #1, opened 2023-12-15
Make locations requirements (half/all) uses actual item count instead of the one from manualitem

nicopop:
As part of my effort to make Manual easier to make work with other player using the same Apworld,
I decided to check the brand new half/all items requirement and saw that it was using the base.item_name_to_item
which wouldnt work correctly if items are removed and thus this draft
Here is the unsent discord message about it:
>currently looking into making the requires:half/all/possible percentage idea I had actually check the real items counts instead of the count from the Manual Item
that way the half/all/etc is accurate to whatever item count a player has vs other players
> 
> my first version of getting the real items counts would add 25ish seconds to gen (for 2 players) so not good
that is if I dont save that data somewhere and only check once per player.
> 
> But I dont know the best practice for this but where should I save processed data since checkRequireStringForArea in rules.py is actually only run when everything in manual is done and is called from multiple different methods 
> 
> list of global enough objects I know I can access in there is: state, base aka World and world aka Multiworld
I tested and I could add myself an atribute dict to all 3 of them

As stated above the location of the attribute can change and the name is also WIP


src/Rules.py
```diff
@@ -63,7 +64,13 @@ def set_rules(base: World, world: MultiWorld, player: int):
...
     def checkRequireStringForArea(state, area):
         requires_list = area["requires"]
-
+        
+        if not hasattr(state, 'manual_item_counts'): #Add a dict so this is only computed once per player, otherwise for a 2 player generation it add 25 seconds of proccess
+            state.manual_item_counts = {}
+        if player not in state.manual_item_counts:
+            real_pool = world.get_items()
+            state.manual_item_counts[player] = {i.name:real_pool.count(i) for i in real_pool if i.player == player}
+        items_counts = state.manual_item_counts[player]
```
nicopop, line 73: this and the 5ish lines above are the important part so generating doesnt take way too long

silasary:
Oh, good catch

FuzzyGamesOn:
Don't forget to move PRs from Draft to Ready for review when you're done getting the code settled. 😄

src/Rules.py
```diff
@@ -77,19 +79,22 @@ def checkRequireStringForArea(state, area):
...
 
             if require_type == 'category':
                 category_items = [item for item in base.item_name_to_item.values() if "category" in item and item_name in item["category"]]
                 if item_count.lower() == 'all':
-                    item_count = sum([base.item_name_to_item[category_item["name"]]['count'] for category_item in category_items])
+                    item_count = sum([items_counts.get(category_item["name"], 0) for category_item in category_items])
                 elif item_count.lower() == 'half':
-                    item_count = sum([base.item_name_to_item[category_item["name"]]['count'] for category_item in category_items]) / 2
+                    item_count = sum([items_counts.get(category_item["name"], 0) for category_item in category_items]) / 2
+                elif item_count.lower().endswith('%') and len(item_count) > 1:
```
FuzzyGamesOn, line 95: Shouldn't need `lower()` here because should just be digits and a percent symbol.

src/Rules.py
```diff
@@ -77,19 +79,22 @@ def checkRequireStringForArea(state, area):
             item_parts = item.split(":")
             item_name = item
             item_count = "1"
-
             if len(item_parts) > 1:
                 item_name = item_parts[0]
                 item_count = item_parts[1]
-
+                
+            item_current_count = items_counts.get(item_name, 0)
```
FuzzyGamesOn, line 86: This isn't used until the block for require_type == 'item', so would move it down there, similar to how category_items is under the == 'category' part.

src/__init__.py
```diff
@@ -275,11 +275,16 @@ def generate_basic(self):
             self.multiworld.itempool.remove(item_to_place)
 
         after_generate_basic(self, self.multiworld, self.player)
-
+        
+        if not hasattr(self, 'manual_item_counts'): #Add a dict so this is only computed once per player, otherwise for a 2 player generation it add 25 seconds of proccess
```
FuzzyGamesOn, line 279: Could just add this as a class prop around line 69 and default it to {}. Also, we don't need to have "manual" in the var name, since it's implied, and the scope is implied to be the current World object.

FuzzyGamesOn (changes requested):

nicopop:
its not perfect but with the fallback it work for both normal and universal tracker

src/Options.py
```diff
@@ -15,6 +15,7 @@ class FillerTrapPercent(Range):
 for category in category_table:
     for option_name in category_table[category].get("yaml_option", []):
         if option_name not in manual_options:
-            manual_options[option_name] = type(option_name, (Toggle,), {"default": True})
+            manual_options[option_name] = type(option_name, (DefaultOnToggle,), {"default": True})
+            manual_options[option_name].__doc__ = "Should items/locations linked to this option be enabled?"
```
nicopop, line 19: small bonus: AP's unit test told me docs are required for all options.
also since the default is true if we want it to be the same on the webhost it need to be a DefaultOnToggle instead

nicopop: Something simple we could do is allow creating off by default options,
by checking for some prefix char like we do for category we could generate options set to false by default.
On my local version I implemented it and tested and it work
```python
for option_name in category_table[category].get("yaml_option", []):
    option_default_off = option_name.startswith('!')
    if option_default_off:
        option_name = option_name[1:]
    if option_name not in manual_options:
        if option_default_off:
            manual_options[option_name] = type(option_name, (Toggle,), {"default": False})
        else: 
            manual_options[option_name] = type(option_name, (DefaultOnToggle,), {"default": True})
        manual_options[option_name].__doc__ = "Should items/locations linked to this option be enabled?"
```

FuzzyGamesOn: I think we should wait to do more with options until we design a proper options.json. The categories stuff aims for a simple way for people to define those with a simple effect, but this is basically the categories problem all over again... in that, to have space to properly allow expanding how people can customize this, we need to separate options into their own json as well.

The prefix stuff was only used in requires because expanding those would've been cumbersome and hard to parse. In general, we should prefer verbose ways to express things, but requires is an exception.

src/Helpers.py
```diff
@@ -14,6 +14,14 @@ def get_option_value(world: MultiWorld, player: int, name: str) -> Union[int, di
...
     return option[player].value
 
+def clamp(value, min, max):
+    if value < min:
+        return min
+    elif value > max:
+        return max
+    else:
+        return value
+
```
FuzzyGamesOn, line 24: HUUUUUGE fan of this, thanks for adding! 😍

src/docs/en_Manual_UltimateMarvelVsCapcom3_Fuzzy.md
nicopop, line 1: oops accidentally pushed this but its still a good way to show apworld maker how they should name this file

src/Helpers.py
```diff
@@ -26,3 +36,45 @@ def is_category_enabled(world: MultiWorld, player: int, category_name: str) -> b
...
+def is_location_enabled(world: MultiWorld, player: int, location: ManualLocation) -> bool:
+    """Check if a location has been disabled by a yaml option."""
+    hook_result = before_is_location_enabled(world, player, location)
+    if hook_result is not None:
+        return hook_result
+
+    return is_manualobject_enabled(world, player, location)
+
+def is_manualobject_enabled(world: MultiWorld, player: int, object: any) -> bool:
```
nicopop, line 72: if this was in c# that method would've been private

src/Helpers.py
```diff
@@ -26,3 +36,45 @@ def is_category_enabled(world: MultiWorld, player: int, category_name: str) -> b
             if not is_option_enabled(world, player, option_name):
                 return False
     return True
+
+def is_item_name_enabled(world: MultiWorld, player: int, item_name: str) -> bool:
```
nicopop, line 40: the location/item_name version wree made to work like is_category_enabled but still call the hook of the base not-name version, if its not something you like they can be removed and the rest will work.

src/Helpers.py
```diff
@@ -26,3 +36,45 @@ def is_category_enabled(world: MultiWorld, player: int, category_name: str) -> b
             if not is_option_enabled(world, player, option_name):
                 return False
     return True
+
+def is_item_name_enabled(world: MultiWorld, player: int, item_name: str) -> bool:
+    """Check if an item named 'item_name' has been disabled by a yaml option."""
+    item = world.worlds[player].item_name_to_item.get(item_name, {})
+    if item == {}:
```
silasary, line 43: I'd do this as `if not item:`

src/__init__.py
```diff
@@ -273,13 +277,17 @@ def generate_basic(self):
...
+            real_pool = self.multiworld.get_items()
+            self.item_counts[self.player] = {i.name:real_pool.count(i) for i in real_pool if i.player == self.player}
 
         after_generate_basic(self, self.multiworld, self.player)
-
         # Uncomment these to generate a diagram of your manual.  Only works on 0.4.4+
 
         # from Utils import visualize_regions
-        # visualize_regions(self.multiworld.get_region("Menu", self.player), f"{self.game}.puml")
+        # visualize_regions(self.multiworld.get_region("Menu", self.player), f"{self.game}_{self.player}.puml")
```
silasary, line 290: I'm not sure this is desirable

nicopop: It help when you gen for multiple users with different removed locations, otherwise it only generate one graph with the last player's options

nicopop: but if unwanted will remove, it was originally acidentally added since the apworld im using to test and program had that

src/Helpers.py
silasary: While python doesn't technically have private methods, PEP8 says "non-public" methods should start with an underscore.

nicopop: Currently reading PEP8 ill make the changes required after.

FuzzyGamesOn:
The breaking change is merged, so you should be good to go Nico 😄

src/__init__.py
```diff
@@ -61,12 +63,18 @@ class ManualWorld(World):
...
+    start_inventory = {}
+
     location_id_to_name = location_id_to_name
     location_name_to_id = location_name_to_id
     location_name_to_location = location_name_to_location
     location_name_groups = location_name_groups
 
+    def interpret_slot_data(self, slot_data: dict[str, any]):
+        #this is called by tools like UT
+        interpret_hooks_slot_data(self, self.player, slot_data)
```
FuzzyGamesOn, line 76: Relatively minor thing, but the hook names all follow the pattern of `*_method_that_has_hook`, where it starts with before or after. In this case, it's not before or after anything, so we should just use a standard word in this case, like hook, but keep the same pattern. So this method should be named something like `hook_interpret_slot_data`.

FuzzyGamesOn (changes requested):
Just the one minor thing. Honestly, super clean stuff otherwise. 💪

FuzzyGamesOn (approved):

PR #2, opened 2023-12-16
Hook templates

axxroytovu:
Starting to add some template/example hooks for people to use as a starting point for their own development


FuzzyGamesOn:
This is a fantastic idea! Originally thought these needed to be comments, but I see now that you're designing the example hooks to start with empty data sets for what the hook is doing. Love that!

Feel free to let me or others know what we can do to help with these. 😄

axxroytovu:
I think these are the major use cases. If you have any other ideas for templates I'm happy to add them.

FuzzyGamesOn:
This is awesome! Love these examples. Thanks for contributing! 😁

PR #3, opened 2023-12-21
Requirement Functions

silasary:
Sometimes you just don't want to type `|5 PLD Levels:7| or |5 WAR Levels:7| or |5 DRK Levels:7| or |5 GNB Levels:7| or |5 WHM Levels:7| or |5 SCH Levels:7| or |5 AST Levels:7| or |5 SGE Levels:7| or |5 MNK Levels:7| or |5 DRG Levels:7| or |5 NIN Levels:7| or |5 SAM Levels:7| or |5 RPR Levels:7| or |5 BRD Levels:7| or |5 MCH Levels:7| or |5 DNC Levels:7| or |5 BLM Levels:7| or |5 SMN Levels:7| or |5 RDM Levels:7| or |5 BLU Levels:7|` hundreds of times.

Also fixes a bug Rooby found when items.json has counts as strings.

### DIFF
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -27,6 +27,8 @@ class DataValidation():
                         # it's just a category, so ignore it
                         if '@' in item:
                             continue
+                        if '$' in item:
+                            continue
 
                         item = item.replace("|", "")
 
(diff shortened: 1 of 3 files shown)


FuzzyGamesOn:
Making a note that we talked before about this looking more like a function; e.g., `hasJobLevels(7)`. Suggested that we might use a different boundary than `||`, which usually just does names/categories/similar, and maybe something like `{}`, but open to anything that makes sense here. Having a separate delimiter for hook-interpreted things is probably good.

This also avoids reusing the `|item name:count|` colon separator for something that isn't a count. Ideally, `:<value>` across Manual would always refer to counts of something. 😃

FuzzyGamesOn (approved):
This is really cool! Super excited for folks to get a chance to mess with this.

Going to wait to merge until the next stable gets out.

PR #4, opened 2023-12-30
Change fillers into a callable function

axxroytovu:
Turning the filler into a function means it can be called from a hook to enable better item handling.


FuzzyGamesOn (approved):
Sure, this seems like good flexibility to have.

PR #6, opened 2024-01-13
Client: Stop trimming colon suffixes from item name lookups

FuzzyGamesOn:
This is based on the (very) old way of defining item counts in Manual, where the `:X` count syntax used for requires now was actually the same way that you defined counts on items in the items.json, too. Since that's no longer the case, we don't need to trim `:X` from the end of item names in our item lookups.

Yes, their requires will still be broken for items with colon in the name. But at least receiving the item at all shouldn't break the client entirely.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -68,17 +68,6 @@ class ManualContext(SuperContext):
 
         self.location_names_to_id = dict([(value, key) for key, value in self.location_names.items()])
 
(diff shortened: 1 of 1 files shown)


silasary:
When updating the Client, you should bump the version here: https://github.com/ManualForArchipelago/Manual/blob/main/src/__init__.py#L366

This will ensure that the newest version of the code always ends up in the launcher.

PR #7, opened 2024-01-22
Breaking change: Moving item creation to the create_items step

FuzzyGamesOn:
This is something that's needed to be adjusted for a while. Currently, Manual creates its item pool in generate_basic, but it really should be created in create_items, since that's what it's there for. generate_basic is more there for post item pool things.

So this PR changes that. Some of the specific changes:
- There are now two hooks to replace before_generate_basic: before_create_items_starting and before_create_items_filler
   - That way, people can hook before whichever process they want during item creation
- The victory location and its event are now setup during create_regions instead of pre_fill, as another "what it's there for" change
- generate_basic now just handles place and don't place.

Updated all the hook templates and created new hooks for create_items, and took item pool out of generate basic hooks.

(As an aside, took out progressive items, since they were never implemented.)

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -6,10 +6,10 @@ from typing import Callable, Optional
 from worlds.generic.Rules import forbid_items_for_player
 from worlds.LauncherComponents import Component, SuffixIdentifier, components, Type, launch_subprocess
 
-from .Data import item_table, progressive_item_table, location_table, region_table, category_table
+from .Data import item_table, location_table, region_table, category_table
 from .Game import game_name, filler_item_name, starting_items
 from .Locations import location_id_to_name, location_name_to_id, location_name_to_location, location_name_groups
-from .Items import item_id_to_name, item_name_to_id, item_name_to_item, advancement_item_names, item_name_groups
+from .Items import item_id_to_name, item_name_to_id, item_name_to_item, item_name_groups
 
 from .Regions import create_regions
 from .Items import ManualItem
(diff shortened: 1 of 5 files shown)


src/__init__.py
```diff
@@ -55,85 +54,71 @@ class ManualWorld(World):
...
-            #     shouldUseProgressive = (self.multiworld.progressive_items[self.player].value);
-
-            #     if shouldUseProgressive and name in self.progressive_item_table:
-            #         name = self.progressive_item_table[name]
+            if name == "__Victory__": continue
+            if name == filler_item_name: continue
 
             item = self.item_name_to_item[name]
-            item_count = 1
+            item_count = int(item.get("count") or 1)
```
silasary, line 94: This will increase `"count": 0` to 1.
You want to use `item.get("count", 1)`.

src/__init__.py
```diff
@@ -313,8 +294,13 @@ def fill_slot_data(self):
 
         return slot_data
     
+    ###
+    # Non-standard AP world methods
+    ###
```
silasary, line 299: If I'm going to be technical, `generate_output` is a standard method

silasary (changes requested):
Looks good, other than a couple minor tweaks

FuzzyGamesOn:
> Looks good, other than a couple minor tweaks

Yep, got those sorted. Thanks for pointing those out! 😁 Forgot about generate_output down there, haha.

nicopop (approved):
looks good, ill need to modify stuff on my pr once this merge but this should work

silasary (approved):

PR #10, opened 2024-01-24
Gracefully handle unknown items and locations in the case of an outdated APWorld and/or multiworld

silasary:
It turns out that CommonClient already knows the names of items and locations from the datapackage, and we were using those names already.

So why not fail softly if the apworld doesn't have an item loaded, and let people connect anyway (with a slightly degraded experience).

This will turn unknown items and locations into fallbacks with a name, but no categories.  It still works, can still send checks, and for the most part, shouldn't be much of a pain point.  Certainly less of a pain point than "Unable to connect, KeyError".


FuzzyGamesOn (approved):
This could potentially mask other issues, but I agree that being more graceful here is better than failing outright with no feedback.

PR #13, opened 2024-01-28
Client: Make client reassign IDs based on datapackage

silasary:
Brazenly empowered by #10, I started doing more aggressive improvements to an in-progress apworld, which exposed new flaws.

This update syncs the network IDs back into your item and location tables, overwriting the ones generated at load time.  Which prevents issues where inserting a new item halfway down your json file causes everything later in the file to take on the wrong name.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -80,7 +80,6 @@ class ManualContext(SuperContext):
         if not location:
             # It is absolutely possible to pull categories from the data_package via self.update_game. I have not done this yet.
             location = AutoWorldRegister.world_types[self.game].location_name_to_location.get(name, {"name": name})
-        location["id"] = self.location_names_to_id.get(name)
         return location
 
     def get_location_by_id(self, id):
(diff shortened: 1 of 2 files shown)


silasary:
I'm not quite sure why, but Locations don't want to be re-id'd.  I'm guessing we set them up differently?

Regardless, this works as intended.

silasary:
Also, we store the last Game string in persistant storage now, because connecting and disconnecting once a minute was being slowed down by needing to type the game string each time 😛

src/ManualClient.py
```diff
@@ -394,8 +407,9 @@ def update_tracker_and_locations_table(self, update_highlights=False):
                                         # Get the item name from the item Label, minus quantity, then do a lookup for count
                                         old_item_text = item.text
                                         item_name = re.sub("\s\(\d+\)$", "", item.text)
+                                        item_id = self.ctx.item_names_to_id[item_name]
                                         item_data = self.ctx.get_item_by_name(item_name)
```
FuzzyGamesOn, line 411: item_data looks to not be used anymore after adding the item_id var.

FuzzyGamesOn (changes requested):

src/ManualClient.py
silasary: Ah, yep. My IDE didn't catch it because we reuse the name later in the method.  Fixed.

PR #14, opened 2024-02-15
Deprecated in 0.4.4 Fix

nicopop:
while testing source with 0.4.4 I got some Deprecation warning and changed what caused those
If I find others ill add to this


FuzzyGamesOn (approved):

nicopop:
Actually if anyone find any other deprecation before this merge you can add to this branch

nicopop:
they deprecated the usage of multiworld.local_items in 0.4.4
you get to use the options one now so any local items will show in the spoilerlog

src/__init__.py
```diff
@@ -115,13 +115,11 @@ def create_items(self):
                 new_item = self.create_item(name)
                 pool.append(new_item)
 
-            if item.get("early") and item.get("local"): # both
-                self.multiworld.local_early_items[self.player][name] = item_count
```
FuzzyGamesOn, line 119: It sounds like local early items is still a thing on the multiworld. Did we want to keep this? @nicopop

nicopop: well it work the same witouth

nicopop: but as a bonus the local items appear in the spoiler log

nicopop: we might need to decide if we prioritize the json over the player yaml, 
that way we could check if the item is in the wrong local/non_local/both and remove it from there or not

nicopop: Something like
```
            if item.get("local", None) != None: # only local
                if item.get("local"):
                    if name not in self.options.local_items.value:
                        self.options.local_items.value.add(name)
                else:
                    if name not in self.options.non_local_items.value:
                        self.options.local_items.value.add(name)
```

FuzzyGamesOn: On principle, YAML will always take precedence over the JSON when they can have contradicting settings. This allows players to customize their easy or painful seed as much as they want. 😃 And if the YAML doesn't generate, they can fix their YAML.

FuzzyGamesOn (approved):

axxroytovu (approved):

PR #16, opened 2024-02-15
Add validation check for bad Starting Items blocks

FuzzyGamesOn:
This adds a couple of checks for starting items, both of which will throw validation errors:

- If the starting item block contains invalid keys
- If the starting item block is not a dict

### DIFF
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -260,6 +260,23 @@ class DataValidation():
                         raise ValidationError("Item category %s is set as a starting item category, but is misspelled or is not defined on any items." % (category_name))
 
     @staticmethod
+    def checkStartingItemsForBadSyntax():
+        if "starting_items" not in DataValidation.game_table:
+            return
+        
+        starting_items = DataValidation.game_table["starting_items"]
+
+        for starting_block in starting_items:
+            if type(starting_block) is not dict or len(starting_block.keys()) == 0:
+                raise ValidationError("One of your starting item definitions is not a valid dictionary.\n   Each definition must be inside {}, as demonstrated in the Manual documentation.")
+                
+            valid_keys = ["items", "item_categories", "random", "if_previous_item"]
+            invalid_keys = ['"{}"'.format(key) for key in starting_block.keys() if key not in valid_keys]
+
+            if len(invalid_keys) > 0:
+                raise ValidationError("One of your starting item definitions is invalid and may have unexpected results.\n   The invalid starting item definition specifies the following incorrect keys: {}".format(", ".join(invalid_keys)))
+
+    @staticmethod
     def checkForGameBeingInvalidJSON():
         if len(DataValidation.game_table) == 0:
             raise ValidationError("No settings were found in your game.json. This likely indicates that your JSON is incorrectly formatted. Use https://jsonlint.com/ to validate your JSON files.")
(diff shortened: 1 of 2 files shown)


src/DataValidation.py
```diff
@@ -259,6 +259,23 @@ def checkStartingItemsForValidItemsAndCategories():
                     if len([item for item in DataValidation.item_table if "category" in item and category_name in item["category"]]) == 0:
                         raise ValidationError("Item category %s is set as a starting item category, but is misspelled or is not defined on any items." % (category_name))
 
+    @staticmethod
+    def checkStartingItemsForBadSyntax():
+        if "starting_items" not in DataValidation.game_table:
+            return
+        
+        starting_items = DataValidation.game_table["starting_items"]
```
axxroytovu, line 267: These can be combined into one line. 
`if not starting_items := DataValidation.game_table.get("starting_items", False):
  return`

Don't know that it's more readable but it is more "pythonic" with the addition of the `:=` operator

src/DataValidation.py
```diff
@@ -259,6 +259,23 @@ def checkStartingItemsForValidItemsAndCategories():
...
+            return
+        
+        starting_items = DataValidation.game_table["starting_items"]
+
+        for starting_block in starting_items:
+            if type(starting_block) is not dict or len(starting_block.keys()) == 0:
+                raise ValidationError("One of your starting item definitions is not a valid dictionary.\n   Each definition must be inside {}, as demonstrated in the Manual documentation.")
+                
+            valid_keys = ["items", "item_categories", "random", "if_previous_item"]
+            invalid_keys = ['"{}"'.format(key) for key in starting_block.keys() if key not in valid_keys]
```
axxroytovu, line 274: F-strings are more stable than `.format`. So `f'"{key}"'` should work.

src/DataValidation.py
```diff
@@ -259,6 +259,23 @@ def checkStartingItemsForValidItemsAndCategories():
...
+        if "starting_items" not in DataValidation.game_table:
+            return
+        
+        starting_items = DataValidation.game_table["starting_items"]
+
+        for starting_block in starting_items:
+            if type(starting_block) is not dict or len(starting_block.keys()) == 0:
+                raise ValidationError("One of your starting item definitions is not a valid dictionary.\n   Each definition must be inside {}, as demonstrated in the Manual documentation.")
+                
+            valid_keys = ["items", "item_categories", "random", "if_previous_item"]
```
nicopop, line 273: looks good But I did mark in schema _comment as a valid key to write comments, with this it would be marked as invalid

axxroytovu (commented):
Both comments are options but not necessary. Code should run fine as is.

nicopop (changes requested):
other than that it looks good
if _comment is not something we want ill instead remove it from the schemas
Next time ill make this a general comment instead of a change required, idk if/how to change an existing review

src/DataValidation.py
FuzzyGamesOn: Good point! I actually completely forgot that the walrus operator was a thing. 😜

FuzzyGamesOn:
> other than that it looks good if _comment is not something we want ill instead remove it from the schemas Next time ill make this a general comment instead of a change required, idk if/how to change an existing review

Nah, I consider it broken if we tell people they can use a schema key and we error when they do. So you were completely in the right asking for a change to support it. 😃

nicopop (approved):

axxroytovu (approved):

PR #23, opened 2024-03-08
Requires functions fix

nicopop:
Function would make generation crash if `||` is used in a function arguments since we reused the raw `area["requires"]`
instead of using `requires_list` that has been modified by the functions

I also Included the functions I made to allow optionnal items, and since its a function the creator must conscientiously choose to use them instead of just a symbol or new item parameters.

if its not something we want tell me and ill revert the comit with those (or delete them if its simpler)

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -332,6 +332,15 @@ class ManualWorld(World):
 
         return item_pool
 
+    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict:
+        """returns the player real item count"""
+        if player == None:
+            player = self.player
+        if player not in self.item_counts or reset:
+            real_pool = self.multiworld.get_items()
+            self.item_counts[player] = {i.name: real_pool.count(i) for i in real_pool if i.player == player}
+        return self.item_counts.get(player)
+
     def client_data(self):
         return {
             "game": self.game,
--- a/src/hooks/Rules.py
+++ b/src/hooks/Rules.py
@@ -1,6 +1,9 @@
+from typing import Optional
 from worlds.AutoWorld import World
+from ..Helpers import clamp
 from BaseClasses import MultiWorld, CollectionState
 
+import re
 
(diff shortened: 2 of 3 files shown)


silasary (dismissed):
Looks good. New functions are very useful

src/hooks/Rules.py
```diff
@@ -24,3 +27,72 @@ def anyClassLevel(world: World, mw: MultiWorld, state: CollectionState, player:
...
+    # parse user written statement into list of each item
+    for item in re.findall(r'\|[^|]+\|', requires):
+        itemScanned = OptOne(base, world, state, player, item, items_counts)
+        requires_list = requires_list.replace(item, itemScanned)
+
+    for function in functions:
+        requires_list = requires_list.replace("{" + function + "(temp)}", "{" + func_name + "(" + functions[func_name] + ")}")
+    return requires_list
+
+def _getItem_counts(base: World, world: MultiWorld, player: int):
```
FuzzyGamesOn, line 93: This function's contents is a duplicate of the same logic in the world's Rules file above: https://github.com/ManualForArchipelago/Manual/pull/23/files#diff-a25c303033861df67335b2a78c209196808d8d93e059588997e9f5422bad35e8L74-L79

Since both places need it, I'd rather see it as a public `get_item_counts` method on the world that needs no arguments, then both places can just call that getter. Or any hooks that might want to call it can call it there too. 😃

FuzzyGamesOn (changes requested):

src/__init__.py
```diff
@@ -332,6 +332,15 @@ def add_filler_items(self, item_pool, traps):
 
         return item_pool
 
+    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict:
+        """returns the player real item count"""
+        if player == None:
```
nicopop, line 337: I left this optionnal player argument for is someone in the futur want to get the counts of another player

FuzzyGamesOn (approved):
Perfect! Optional player sounds great too.

PR #27, opened 2024-03-09
New regions.json key: `entrance_rules`.  Allows you to make entering a region cost different things based on where you're coming from.

silasary:
For example, requiring different baits for a fish depending on where you're fishing.

There is no cost to using `add_rule` instead of `set_rule`, as AP treats it like `set_rule` the first time you call it on an entrance.

If you have both a requires and an entrance rule, it'll be a boolean AND.


FuzzyGamesOn:
The code looks good. Worth noting that this clashes somewhat with the `connects_to` setting on regions (in terms of affecting entrances vs. exits), which could be a source of confusion for world creators trying to grasp how their settings affect things. 

However, I think it's overall fine if we mix and match these, as long as we're explicit about which direction they're configuring -- like how `connects_to` is. So maybe a key like `rules_from` would work better, so they get used to seeing `to` and `from` to indicate which direction (entrance / exit) it affects.

silasary:
> Worth noting that this clashes somewhat with the connects_to setting on regions (in terms of affecting entrances vs. exits
Yeah, I was just thinking about that.

silasary:
Doing a bit of an overhaul to make it feel better

silasary:
Finally reviving this one

schemas/Manual.regions.schema.json
```diff
@@ -39,6 +39,32 @@
                             "type":"boolean",
                             "default": false
                         },
+                        "exit_requires": {
+                            "description": "(Optional) Additional requirements to use a specific exit from this region.",
```
FuzzyGamesOn, line 43: We should make these descriptions more plain and have them incorporate language that links them to `connects_to`, since they're gating that setting. Something like "to exit this region to a connected region."

And for the entrance_requires, something like "to enter this region from a region that connects to this one." Tweak as needed.

FuzzyGamesOn (changes requested):
Minor text change for schema desc, rest looks fine 😃

nicopop (approved):

FuzzyGamesOn (approved):

PR #28, opened 2024-03-16
Fix the OptOne function and Comment Typo

nicopop:
It seems I forgot to test using |item:Count| after I optimised the optOne function 
I also fixed the typo Hallstead found in the comments

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -341,11 +341,11 @@ class ManualWorld(World):
 
         return item_pool
 
-    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict:
+    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict[str, int]:
         """returns the player real item count"""
         if player == None:
             player = self.player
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -341,11 +341,11 @@ def add_filler_items(self, item_pool, traps):
 
         return item_pool
 
-    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict:
+    def get_item_counts(self, player: Optional[int] = None, reset: bool = False) -> dict[str, int]:
         """returns the player real item count"""
         if player == None:
             player = self.player
-        if player not in self.item_counts or reset:
+        if not self.item_counts.get(player, {}) or reset:
```
nicopop, line 348: This change was made so if creator call get_item_countsbefore the items are created it doesnt stop the count from generating later when required

silasary: Fun

silasary (approved):
Looks good to me

FuzzyGamesOn (approved):

PR #29, opened 2024-03-16
Automatically remove excess filler if there's too much

silasary:
This is no substitute for making a properly sized world.  But it's better than what we're currently doing

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -344,7 +344,7 @@ class ManualWorld(World):
                 extra_item = self.create_item(filler_item_name)
                 item_pool.append(extra_item)
         elif extras < 0:
-            logging.warning(f"{self.game} has more items than locations. {abs(extras)} useless items will be removed at random.")
+            logging.warning(f"{self.game} has more items than locations. {abs(extras)} non-progression items will be removed at random.")
             fillers = [item for item in item_pool if item.classification == ItemClassification.filler]
             traps = [item for item in item_pool if item.classification == ItemClassification.trap]
             useful = [item for item in item_pool if item.classification == ItemClassification.useful]
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -169,7 +169,7 @@ class ManualWorld(World):
         self.start_inventory = {i.name: items_started.count(i) for i in items_started}
 
         pool = before_create_items_filler(pool, self, self.multiworld, self.player)
-        pool = self.add_filler_items(pool, traps)
+        pool = self.adjust_filler_items(pool, traps)
         pool = after_create_items(pool, self, self.multiworld, self.player)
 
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -331,13 +335,33 @@ def add_filler_items(self, item_pool, traps):
...
+            for _ in range(0, trap_count):
                 extra_item = self.create_item(self.random.choice(traps))
                 item_pool.append(extra_item)
 
-            for i in range(0, filler_count):
+            for _ in range(0, filler_count):
                 extra_item = self.create_item(filler_item_name)
                 item_pool.append(extra_item)
+        elif extras < 0:
+            print(f"Warning: {self.game} has more items than locations. {abs(extras)} useless items will be removed at random.")
```
nicopop, line 346: maybe use a logger instead of print so it shows in log?

silasary: Oh, yes. I definitely meant to do that.

src/__init__.py
```diff
@@ -331,13 +336,33 @@ def add_filler_items(self, item_pool, traps):
...
+            for _ in range(0, abs(extras)):
+                popped = None
+                if fillers:
+                    popped = fillers.pop()
+                elif traps:
+                    popped = traps.pop()
+                elif useful:
+                    popped = useful.pop()
+                else:
+                    print("Warning: Could not remove enough non-progression items from the pool.")
```
nicopop, line 363: this one too ;)

nicopop:
dont forget to change the call to the self.add_filler_items at line 172

nicopop (changes requested):
Since you deprecated add_filler_items you also need to change line 172 of \_\_init__.py or it will get called when generating.

silasary:
Did I seriously forget to stage that line?

nicopop (dismissed):
Looks good now 👍

src/__init__.py
```diff
@@ -331,13 +336,33 @@ def add_filler_items(self, item_pool, traps):
...
+            for _ in range(0, trap_count):
                 extra_item = self.create_item(self.random.choice(traps))
                 item_pool.append(extra_item)
 
-            for i in range(0, filler_count):
+            for _ in range(0, filler_count):
                 extra_item = self.create_item(filler_item_name)
                 item_pool.append(extra_item)
+        elif extras < 0:
+            logging.warning(f"{self.game} has more items than locations. {abs(extras)} useless items will be removed at random.")
```
FuzzyGamesOn, line 347: I think calling the items useless when useful is included could be misleading. We should change this to "non-progression".

FuzzyGamesOn (changes requested):
Just a minor verbiage change for clarity 😄

FuzzyGamesOn (approved):

PR #32, opened 2024-03-19
Renamed world: Multiworld to multiworld: Multiworld

nicopop:
and also renamed base: World to world: World


silasary (approved):
Looks great

FuzzyGamesOn (approved):

PR #34, opened 2024-03-21
Add the foundation for UT integrated hooks

axxroytovu:
Adds a "hook" that can enable the universal tracker secondary generation.

Some games are using hooks to randomize the location set, and this does not interact well with how Universal Tracker shows available locations.

This foundation allows the author to know if the generation is a UT gen or not, and can modify their hooks accordingly.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -35,7 +35,6 @@ try:
     from .hooks.World import ut_enabled
 except Exception as e:
     ut_enabled = False
-    print(e)
 
 
 class ManualWeb(WebWorld):
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -31,6 +31,11 @@ from .hooks.World import \
     before_generate_basic, after_generate_basic, \
     before_fill_slot_data, after_fill_slot_data, before_write_spoiler
 from .hooks.Data import hook_interpret_slot_data
+try:
+    from .hooks.World import ut_enabled
+except Exception as e:
+    ut_enabled = False
+    print(e)
 
 
 class ManualWeb(WebWorld):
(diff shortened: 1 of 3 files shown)


src/__init__.py
```diff
@@ -386,6 +390,15 @@ def client_data(self):
...
         }
+    
+    # for the universal tracker, doesn't get called in standard gen
+    @staticmethod
+    def return_slot_data(slot_data):
+        # returning slot_data so it regens, giving it back in multiworld.re_gen_passthrough
+        return slot_data
+    
+    if ut_enabled:
+        interpret_slot_data = return_slot_data
```
silasary, line 401: It's worth noting that we already have a `interpret_slot_data` at https://github.com/ManualForArchipelago/Manual/pull/34/files#diff-6b39a2f1a6399aa39e427fb4c11a7b552e615a2d83a3c36aa6dd6ddbbb8281f4R82-R84 and a hook for it in hooks/data.py (Why is it there??)

Also, this somehow conflicts with #35

axxroytovu: Ok I fixed this to use the correct interpret_slot_data function

I also removed the weird hook that doesn't do anything

axxroytovu:
Universal Tracker regenerates the world if it sees that the world has the `interpret_slot_data` method, regardless of what the method actually returns. The way this PR is set up, the world won't show as having the method unless you change the boolean to True.

FuzzyGamesOn:
> Universal Tracker regenerates the world if it sees that the world has the `interpret_slot_data` method, regardless of what the method actually returns. The way this PR is set up, the world won't show as having the method unless you change the boolean to True.

Looks like it doesn't regenerate if the interpret_slot_data on the world returns `None`: https://github.com/FarisTheAncient/Archipelago/blob/Tracker_v0.1.4/worlds/tracker/TrackerClient.py#L186-L190

Seems like it would be better to have the hook method return None by default, which world creators can override if they want. Then we just return the results of the hook method in interpret_slot_data. Since that hook is the only way for world creators to interact with interpret_slot_data, too.

The tip on detecting the UT gen in the hook is handy, though, would keep. 😃

axxroytovu:
Ok. I'll add the tip to the docs and close this PR

PR #35, opened 2024-03-21
Automatically store yaml options in slot data, and restore it for UT

silasary:


src/__init__.py
```diff
@@ -77,7 +78,15 @@ class ManualWorld(World):
...
-        hook_interpret_slot_data(self, self.player, slot_data)
+
+        regen = False
+        for key, value in slot_data.items():
+            if key in self.options_dataclass.type_hints:
+                getattr(self.options, key).value = value
+                regen = True
+
+        regen = hook_interpret_slot_data(self, self.player, slot_data) or regen
+        return regen
```
FuzzyGamesOn, line 89: Mentioned in axxroy's PR but, for some reason, UT only treats `None` as a value to not regenerate a world. So both True and False here would regen. Definitely something that they should fix, but we can use None / True in the meantime probably.

silasary: Easy.  I also realized from reading that PR that I should also be comparing yaml setting against Slot Data, and only flagging if they differ.

silasary: Sent PR to Tracker, local fix coming shortly
https://github.com/FarisTheAncient/Archipelago/pull/2

PR #36, opened 2024-04-04
Adding Location events

FarisTheAncient:
Adds location events to every location, that can be used in place of any item

used by surrounding the location name in |[ ]|

### DIFF
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -33,6 +33,10 @@ class DataValidation():
                         item_parts = item.split(":")
                         item_name = item
 
+                        # it's just a location, so ignore it
+                        if re.match(r'^\[.*\]$',item_name):
+                            continue
+
                         if len(item_parts) > 1:
                             item_name = item_parts[0]
 
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -34,15 +34,14 @@ class DataValidation():
                         item_parts = item.split(":")
                         item_name = item
 
-                        # it's just a location, so ignore it
+                        if len(item_parts) > 1:
+                            item_name = item_parts[0]
+
                         if re.match(r'^\[.*\]$',item_name):
                             loc_name = item_name.replace("[","").replace("]","")
                             if loc_name not in DataValidation.location_names:
                                 raise ValidationError("Location %s is required by location %s but is misspelled or does not exist." % (loc_name, location["name"]))
                             continue
-
-                        if len(item_parts) > 1:
-                            item_name = item_parts[0]
 
                         item_exists = len([item["name"] for item in DataValidation.item_table if item["name"] == item_name]) > 0
 
@@ -65,6 +64,12 @@ class DataValidation():
                             if len(or_item_parts) > 1:
                                 or_item_name = or_item_parts[0]
 
+                            if re.match(r'^\[.*\]$',or_item_name):
+                                loc_name = or_item_name.replace("[","").replace("]","")
+                                if loc_name not in DataValidation.location_names:
+                                    raise ValidationError("Location %s is required by location %s but is misspelled or does not exist." % (or_item_name, location["name"]))
+                                continue
+
                             item_exists = len([item["name"] for item in DataValidation.item_table if item["name"] == or_item_name]) > 0
 
                             if not item_exists:
@@ -75,6 +80,12 @@ class DataValidation():
 
                         if len(item_parts) > 1:
                             item_name = item_parts[0]
+
+                        if re.match(r'^\[.*\]$',item_name):
+                            loc_name = item_name.replace("[","").replace("]","")
+                            if loc_name not in DataValidation.location_names:
+                                raise ValidationError("Location %s is required by location %s but is misspelled or does not exist." % (loc_name, location["name"]))
+                            continue
 
                         item_exists = len([item["name"] for item in DataValidation.item_table if item["name"] == item_name]) > 0
 
@@ -107,6 +118,12 @@ class DataValidation():
                         if len(item_parts) > 1:
                             item_name = item_parts[0]
 
+                        if re.match(r'^\[.*\]$',item_name):
+                            loc_name = item_name.replace("[","").replace("]","")
+                            if loc_name not in DataValidation.location_names:
+                                raise ValidationError("Location %s is required by region %s but is misspelled or does not exist." % (loc_name, region_name))
+                            continue
+
                         item_exists = len([item["name"] for item in DataValidation.item_table if item["name"] == item_name]) > 0
 
                         if not item_exists:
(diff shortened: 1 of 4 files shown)


src/Regions.py
```diff
@@ -61,7 +62,11 @@ def create_region(world: World, multiworld: MultiWorld, player: int, name: str,
         for location in locations:
             loc_id = world.location_name_to_id.get(location, 0)
             locationObj = ManualLocation(player, location, loc_id, ret)
+            eventLocationObj = ManualLocation(player,"["+location+"]",None,ret)
```
nicopop, line 65: Idk if we want to create a event loc and item for each location, might want to check for a new attribute in locations something like "create_event"

nicopop (changes requested):
the first comment is really the requested changes the other is more of a needed discussion.
also I made those comment before testing if this work or not ;)

src/Regions.py
nicopop: actually now that I gened a game it does double the locations in the spoiler logs so its probably a good idea to only add those if required

src/DataValidation.py
```diff
@@ -36,6 +37,12 @@ def checkItemNamesInLocationRequires():
                         if len(item_parts) > 1:
                             item_name = item_parts[0]
 
+                        if re.match(r'^\[.*\]$',item_name):
```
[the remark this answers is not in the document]
nicopop: actually we probably should do that to the rest of the validation but thats not on you so you can consider this one as optionnal

src/Regions.py
nicopop: actually I found out you can hide locations from the spoiler list of location by using `eventLocationObj.show_in_spoiler = False`
it still show you getting the event when required in the log but it doesnt double the location names in the location list

FarisTheAncient: I wonder if it would make sense to not have the flag be on locations, but to build the list of events we need to create as we're validating the json's

FarisTheAncient: 100% the events shouldn't be in the list, i can mark them as hidden, if i worked more on the worlds side of ap i'd have caught that myself

src/DataValidation.py
```diff
@@ -9,6 +9,8 @@ class DataValidation():
     game_table = {}
     item_table = []
     location_table = []
+    location_events_table = []
```
FuzzyGamesOn, line 12: This doesn't belong in DataValidation. You're not validating anything with it. 

If you need it to be shared for multiple places, it should go on the World itself. And you can build the list in the Regions file.

src/Regions.py
```diff
@@ -62,6 +64,12 @@ def create_region(world: World, multiworld: MultiWorld, player: int, name: str,
             loc_id = world.location_name_to_id.get(location, 0)
             locationObj = ManualLocation(player, location, loc_id, ret)
             ret.locations.append(locationObj)
+            if location in DataValidation.location_events_table:
+                eventLocationObj = ManualLocation(player,"["+location+"]",None,ret)
+                eventLocationObj.show_in_spoiler = False
+                eventItemOjb = ManualItem("["+location+"]",ItemClassification.progression, None, player)
```
FuzzyGamesOn, line 70: I think this explains why you wanted to go with `|[whatever]|` for the format in requires. But we don't want to present locations to the user as "special items". We want to present them as locations, as their own thing. This is consistent with how we present requirement functions differently, too.

For example, if we use this in a requires string:
```
"requires": "|Key| and |Shield| and [Saved Zelda in the Castle]"
```
... then it is visually clear that "Saved Zelda in the Castle" isn't an item. Because, to the user / world creator, it isn't. It's a location. We don't need to visually represent it in the same way that we have to for making it functional. 

In the background, we can spice it up with extra `[]` or `__LOCATION__:` or whatever we want the format of the event to be. So that `Saved Zelda in the Castle` actually matches `THISISALOCATION:Saved Zelda in the Castle` or whatever.

For an example of how we'd handle the `[]` delimiter like `||`, check out Rules.py around line 82 or so. That's the implementation for handling the `{}` delimiter for requirement functions, and you could probably borrow a lot of what it's doing.

FuzzyGamesOn (changes requested):

FuzzyGamesOn:
I don't see any Manual client changes in this PR. Want to make sure that the event locations / event items don't show up in the client. If they do, this PR will need to adjust that too. 😃

src/Regions.py
FarisTheAncient: The problem is that as an event, as far as AP is concerned it **IS** just an item, the only difference is it's an item that doesn't get sent to the user, or even live beyond the scope of generation

Like i can make the change so || is validated as an item, and [] is validated as a location, but under the hood they're both turned into item access rules, if that satisfies your requirement that they look different to builders and players

FuzzyGamesOn: Yep, exactly. That change is what I'm suggesting. 😃 

Manual was originally designed for -- and still caters to a lot of -- less technical world creators. A lot of these users will never see the internals of AP (or even a custom apworld), so they won't see the correlation. To them, they just see (via Manual) a locations.json and an items.json and fill in stuff, so the two have separate meaning for them. Manual already changes the presentation from the implementation in some cases to reduce complexity for the user, so that's okay (and desirable) to do here, too. Also avoids an event discussion that doesn't need to happen with the user, because it's not information they need. 😃

src/DataValidation.py
FuzzyGamesOn: I see what you're saying. You need to read the event location names from the requires to know which event locations to create.

Problem with putting them in DataValidation is that it shouldn't do anything atomic. Shouldn't store anything for reuse, shouldn't construct anything, change anything... It just should look at the data that's there and detect issues. We don't want to end up tracking down issues in the validation causing bugs in generation, because that would be kinda cursed. 😃 Also, if we ever make validation optional, it shouldn't break generation.

But it's okay to just use list comprehension on the location_table to get locations that have event locations, then iterate those in a new loop. Can't imagine that would have much impact, and makes more sense to move event location creation / prep closer to Regions like you wanted to.

In fact, since it's going to end up on the world object, I could also see the list happening in Data.py and getting imported by the world. Then, in the future, we could run validation on the event location names if we wanted to. But that's not necessary for this. Just somewhere more appropriate than DataValidation, and a new loop is fine. 😃

FarisTheAncient: your validation is already optional if you generate with skip output 😉

FarisTheAncient:
> I don't see any Manual client changes in this PR. Want to make sure that the event locations / event items don't show up in the client. If they do, this PR will need to adjust that too. 😃

I did check and both generate, spoiler log, and Universal Tracker all handled the event locations correctly

src/DataValidation.py
FuzzyGamesOn: Why tho? You went from using an existing loop in DataValidation that was doing something else but also did what you wanted, to not using a loop at all and just creating unnecessary event locations. I'm not understanding why you can't just make that same loop again in one of the proper places.

nicopop:
Im using a slightly modified version of this in my OuterWilds Manual the only change I did was rename the `__LOCATION_` thingy to `[Event] ` so its both prettier in the spoiler logs and easier to add_rule said event item

FuzzyGamesOn:
This PR hasn't seen any activity in over 6 months, so closing. My last comments on this detail the changes we'd need before considering this change again.

PR #37, opened 2024-04-05
Add Support for changing the documentation via game.json and/or hook

nicopop:
I know im probably the only one that make proper doc for my manual apworlds but I wanted to officially make some hook so I dont have to make sure I dont override my changes in __init__.py and lose my custom documentation stuff with any major update.

It need to be this early in the code because from my testings, webhost doesnt execute anything other than whats required to get the data, in this case __doc__ = set_world_doc(__doc__) and web = set_world_webworld(ManualWeb())
~~I placed the docs  in game.json but it could easyly get moved somewhere else~~

edit:
I moved the docs stuff to meta.json/~~meta.py~~data.py and also added a property to enable/disable the puml generation from there 
So there's no real reason for someone making hooks to modify \_\_init\_\_.py and thus can override the base files to update in the future


nicopop:
I want to add schemas before I make this a full pr

FuzzyGamesOn:
Love the idea. My initial thought is that we might want doc-related things separate from the functional pieces in game.json, since it functions more like a global functionality file. So maybe a meta json or a docs json or something. Can see what other folks think 😃 

Definitely a useful feature to have tho.

src/hooks/Meta.py
```diff
@@ -0,0 +1,18 @@
...
+# Return a string containing a custom world description
+# for more info check https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/world%20api.md#docstrings
+def hook_set_world_description(base_doc: str) -> str:
+    return base_doc
+
+
+# this hooks allows you to modify the Webworld that get shown on the WebHost when hosted locally
+# for more info check https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/world%20api.md#webworld-class
+def hook_set_world_webworld(web: WebWorld) -> WebWorld:
+    return web
```
silasary, line 18: Both these hooks feel like they should be in Data.

FuzzyGamesOn: I think this feels over-engineered a bit. The other json files just have a before processed hook, and that makes sense. An after load file hook in Data.py for meta seems good, too, for consistency. Beyond that, I think we can expect people to construct the contents like they would have to with other files. Probably not as much need for individual hooks to target portions of the JSON data.

In the general sense, I see hooks as giving world creators programmatic control over the data broadly, and as a way for them to hook into individual generation steps. Beyond that, in most cases, we're probably getting too granular. 😃

src/__init__.py
```diff
@@ -294,10 +296,11 @@ def generate_basic(self):
 
 
         after_generate_basic(self, self.multiworld, self.player)
-        # Uncomment these to generate a diagram of your manual.  Only works on 0.4.4+
 
-        # from Utils import visualize_regions
-        # visualize_regions(self.multiworld.get_region("Menu", self.player), f"{self.game}_{self.player}.puml")
+        # Enable this in Meta.json to generate a diagram of your manual.  Only works on 0.4.4+
+        if enable_puml:
```
FuzzyGamesOn, line 301: I love this, because I've definitely found myself wanting an option I could turn on to make the UML. I'm torn on whether it should be a YAML option or a JSON setting, though.

Might name it something more obvious like "region_diagram" instead of "puml", for the less technical folks in the community. 😃

src/hooks/Meta.py
nicopop: I tried to have only one hooks  but the code thats is executed by webhost doesnt support self since its undefine at that moment otherwise I would've done 1 hook that allow to change all the data in one go,
thats why im pluggin myself into the variable with a function

src/__init__.py
nicopop: I moved it into meta.py so a hook could change it if someone want to enable but that way.
ill rename it quickly.

src/hooks/Meta.py
nicopop: I might've found a way to combine the hooks into one, ill talk about it in discord

FuzzyGamesOn: Oh, I was just talking about a hook like the other ones used with the JSON files. Like this: https://github.com/ManualForArchipelago/Manual/blob/main/src/Locations.py#L6

src/data/meta.json
```diff
@@ -0,0 +1,37 @@
+{
+    "_comment": "remove the _ before a non-comment/info propertie to enable it",
+    "_comment2": "For more info use json schemas to get a description and allowed value for every properties",
+    "_comment3": "https://github.com/ManualForArchipelago/Manual/blob/main/docs/resources/json-schemas-for-world-devs.md",
```
FuzzyGamesOn, line 4: I feel like we should handle multi-line comments differently. Like having an array or something. Like apworld_description is doing.

nicopop:
in the future ill probably experiment in the schema branch (ill create it again)
to try and make a 'global' schema for defining things like comments so I only need to change it in 1 file

src/Meta.py
```diff
@@ -0,0 +1,48 @@
...
+from worlds.AutoWorld import World, WebWorld
+from .hooks.Data import hook_early_modify_world_properties
+from .Data import meta_table
+
+enable_region_diagram = meta_table.get("enable_region_diagram", False)
+
+def set_world_attributes(world_properties: dict) -> dict:
+    world_properties['description'] = set_world_doc(world_properties.get('description'))
+    world_properties['webworld'] = set_world_webworld(world_properties.get('webworld'))
+    world_properties = hook_early_modify_world_properties(world_properties)
```
FuzzyGamesOn, line 12: I think this hook, and the method it's in, are a source of confusion for me, and my other comments were kinda getting at this. 

So, with the new meta.json, world creators can define content for their WebWorld. That makes sense and is awesome 😃  But what's the use case for having a hook for these properties?

Reasons I ask:
- Hooks are usually a way for a world creator to interact with steps in generation that they don't typically have access to. But meta.json is something they have access to fully, and there's already a hook for after the meta file is loaded in Data. (And other json files have a `before_whatever` hook in their separate files, so that'd make sense for Meta too, even if redundant.) So they already have the ability to define it upfront and to modify it programmatically before it's used.
- A use case for hooks is often getting access to options to dynamically change values based on the option values. But `set_world_attributes` is called before generation starts, so that's not a use case here.

So, without the two use cases above, I'm wondering if we need this hook.

And if we don't need this hook here, we don't need a set world attributes function to dynamically set props on World, right? This file could just format things with newlines and what not, and give the world doc and webworld back to init like Locations gives location_id_to_name or Items gives item_id_to_item or similar, via the import in init. 😃

nicopop: I added a a hook because some people might want to use the python classes and do some minimal dynamic changes
I personally used it to include the version of my manual but now that you remind me that I added the 'after_load_meta_file' that could do just that ill probably remove the hook and do like you said

nicopop: I could convert set_word_doc and set_worlds_webworld to act like location_id_to_name but I would need to move the webworld def and doc to meta.py from \_\_init\_\_.py

FuzzyGamesOn: Yeah, functions in the file are fine for organizing. Mainly just having the file create the values for import, which it does now. 😃 

Changes look great! Very clean.

src/Meta.py
```diff
@@ -0,0 +1,71 @@
...
+#################
+world_description: str = set_world_description("""
+    Manual games allow you to set custom check locations and custom item names that will be rolled into a multiworld.
+    This allows any variety of game -- PC, console, board games, Microsoft Word memes... really anything -- to be part of a multiworld randomizer.
+    The key component to including these games is some level of manual restriction. Since the items are not actually withheld from the player,
+    the player must manually refrain from using these gathered items until the tracker shows that they have been acquired or sent.
+    """)
+world_webworld: ManualWeb = set_world_webworld(ManualWeb())
+
+enable_region_diagram = meta_table.get("enable_region_diagram", False)
```
FuzzyGamesOn, line 71: I'd probably create an entry in meta.json with this set to false as well. That way, users know it's something they can toggle without having to reference docs or something.

That's all I noticed, though. Once that's done, this should be good to go!

src/data/meta.json
```diff
@@ -0,0 +1,40 @@
...
+            ],
+            "_options_presets": {
+                "Easy":{
+                    "Option1": true,
+                    "Option2": "potato"
+                }
+            }
+        }
+    },
+    "_Comment":"Enable the generation of puml diagram of your apworld region and locations for debug purposes",
```
FuzzyGamesOn, line 38: I hate to be that guy, but this is capitalized.

nicopop: cant have 2 of the same tag in json or you get a warning but ill rename it to something else like _comment_for_enable_region_diagram or _info

nicopop: I went with _comment_ instead to keep it clean

FuzzyGamesOn (approved):

PR #38, opened 2024-04-09
Write documentation for Downloading and Playing a Manual.

silasary:
Cleaned up existing "Playing a Manual" docs, and put it here

I also removed placeholder for "Download Client", because it's not really needed.


FuzzyGamesOn (approved):
Looks great! Thanks for knocking that out 😁

PR #39, opened 2024-04-10
Add logic for item with values

nicopop:
I made a function for the [Logic Based on Arbitrary Item Value instead of Discrete Count](https://discord.com/channels/1097532591650910289/1207434345149767710)
and I want to make it official.
At first it was 1 rule function and 1 basic function but I decided to include it in the base manual code
and create some usefull helper functions
I decided to leave it as a rule function because if I add another symbol we'll run out of those quickly

Like my Doc pr Im making it a draft to start because I still got to add it in schemas and also some data validation because I think if a item value type is used then the items with it should all be progressions


src/Helpers.py
```diff
@@ -85,4 +87,32 @@ def _is_manualobject_enabled(multiworld: MultiWorld, player: int, object: any) -
             enabled = False
             break
 
-    return enabled
\ No newline at end of file
+    return enabled
+
+def get_items_for_player(multiworld: MultiWorld, player: int) -> List[Item]:
+    """Return list of items of a player including placed items"""
+    return [i for i in multiworld.get_items() if i.player == player]
```
FuzzyGamesOn, line 94: Huh, I guess I never noticed, but it's weird that `get_locations` takes an optional player id but `get_items` doesn't. 😜

nicopop:
> This seems cool so far! I know it's still in draft, but wanted to take a peek. 😃 

When I make a draft, I'm always open for comments and suggestions. If I don't want any, I'll wait until im ready and skip the draft phase.

nicopop:
ill wait for #37 to be merge before adding the schema to this one so I dont have to deal with merge conflict

nicopop:
im making it a normal pr so you can just tell me what you think of the code/way I did it

src/Helpers.py
```diff
@@ -85,4 +87,37 @@ def _is_manualobject_enabled(multiworld: MultiWorld, player: int, object: any) -
...
+    Keep a cache of the result and wont redo unless 'force == True'
+    """
+    if player is None:
+        player = world.player
+
+    player_items = get_items_for_player(multiworld, player)
+    # Just a small check to prevent caching {} if items don't exist yet
+    if not player_items:
+        #raise Exception("get_items_with_value shouldn't be called before items are created.")
+        return {value: -1}
```
nicopop, line 108: idk if its better to raise Exception here or return empty dict when this is used too early in by a dev

src/DataValidation.py
```diff
@@ -170,6 +173,11 @@ def checkItemsThatShouldBeRequired():
                 if isinstance(location_requires, str):
                     if '|{}|'.format(item["name"]) in location_requires:
                         raise ValidationError("Item %s is required by location %s, but the item is not marked as progression." % (item["name"], location["name"]))
+                    if itemValues and 'ItemValue' in location_requires:
+                        for value in re.findall(r'\{ItemValue\(([^:]*)\:[^)]+\)\}', location_requires):
+                            if value.lower().strip() in itemValues:
+                                raise ValidationError(f"Item {item['name']} has a value for the '{value}' value required by location '{location['name']}', but the item is not marked as progression.")
```
silasary, line 179: Not sure how I feel about this.  I'd much rather rely entirely on `checkIfEnoughItemsForValue` for checking there's enough progression value.  

Right now, the main thing this validation does is say "5 Rupees" cannot be a filler item.

src/DataValidation.py
```diff
@@ -188,10 +196,68 @@ def checkItemsThatShouldBeRequired():
...
                     if item["name"] in region_requires:
                         raise ValidationError("Item %s is required by region %s, but the item is not marked as progression." % (item["name"], region_name))
 
+    @staticmethod
+    def checkIfEnoughItemsForValue():
+        values_available = {}
+        values_requested = {}
+
+        # get all the available values with total count
+        for item in DataValidation.item_table:
```
silasary, line 214: Throw a `if progression or skip balancing` here instead of the above checks, so only progression items are counted towards this check.

silasary (changes requested):

src/DataValidation.py
nicopop: that a solution I didnt consider before checking if there is enough progression and the rest is just bonus value.

src/DataValidation.py
```diff
@@ -192,6 +192,67 @@ def checkItemsThatShouldBeRequired():
...
+            # if the item is already progression, no need to check
+            if not "progression" in item or not item["progression"]:
+                continue
+
+            # progression_skip_balancing is also progression, so no check needed
+            if not "progression_skip_balancing" in item or not item["progression_skip_balancing"]:
+                continue
```
silasary, line 208: Items either have one of the other.  You're eliminating both independently, so no items will make it past the second check.

Replace both of these with the following:
```py
if not item.get("progression") and not item.get("progression_skip_balancing"):
    continue
```

nicopop: I went with the De Morgan law of boolan logic to invert it, I'll modify it later when I got time to test it properly

src/DataValidation.py
```diff
@@ -192,6 +192,67 @@ def checkItemsThatShouldBeRequired():
...
+                    if not values_requested.get(value):
+                        values_requested[value] = count
+                    else:
+                        values_requested[value] = max(values_requested[value], count)
+
+        # compare whats available vs requested
+        for value, count in values_requested:
+            if values_available.get(value, 0) < count:
+                raise ValidationError(f"there's not enough progression items of value '{value}' to fulfill the required {count} {value}. Only a total worth of {values_available.get(value, 0)} {value} can be found.")
+
```
FuzzyGamesOn, line 255: Loving the validation here! Should help reduce the complexity a bit for users 😃

src/Helpers.py
FuzzyGamesOn: This only triggers when the user has no items defined, so I think it's okay to just return a <1 value. If they try to figure out why their item values aren't working, they should figure out that they have no items pretty quickly 😜

FuzzyGamesOn (changes requested):
We should include examples of how to use this feature in the example apworld.

src/DataValidation.py
```diff
@@ -192,6 +195,110 @@ def checkItemsThatShouldBeRequired():
...
+            errors = []
+            for value, count in values_requested.items():
+                if values_available.get(value, 0) < count:
+                    errors.append(f"       '{value}': {values_available.get(value, 0)} out of the {count} {value} worth of progression items required can be found.")
+            if errors:
+                raise ValidationError("There are not enough progression items for the following values: \n" + "\n".join(errors))
+
+    @staticmethod
+    def preFillCheckIfEnoughItemsForValue(world: World, multiworld: MultiWorld):
+        from .Helpers import get_items_with_value, get_items_for_player
```
nicopop, line 263: if I import that at the top it cause a circular import.

silasary: That's fine

src/DataValidation.py
```diff
@@ -338,7 +444,16 @@ def checkForNonStartingRegionsThatAreUnreachable():
...
+def runPreFillDataValidation(world: World, multiworld: MultiWorld):
+    validation_errors = []
 
+    # check if there is enough items with values
+    try: DataValidation.preFillCheckIfEnoughItemsForValue(world, multiworld)
+    except ValidationError as e: validation_errors.append(e)
+
+    if validation_errors:
+        newline = "\n"
+        raise Exception(f"\nValidationError(s) for pre_fill of player {world.player}: \n\n{newline.join([' - ' + str(validation_error) for validation_error in validation_errors])}\n\n")
```
nicopop, line 456: f string doesn't like it when you have a backslash in the {} but its ok if its in variable...

src/DataValidation.py
```diff
@@ -192,6 +195,109 @@ def checkItemsThatShouldBeRequired():
...
+                continue
+
+            Manualregion = DataValidation.region_table.get(region.name, {})
+            if "requires" in Manualregion and Manualregion["requires"]:
+                region_requires = json.dumps(Manualregion["requires"])
+
+                DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, region_requires)
+
+            for location in region.locations:
+                ManualLocation = world.location_name_to_location.get(location.name, {})
```
FuzzyGamesOn, line 278: ManualLocation is a class in Locations.py. I know that's not imported here, but would prefer to avoid the ambiguity. Variable names shouldn't be capitalized, anyways, since that's usually only for class names... so should adjust the region one above, too.

nicopop: i'll be honest idk why I used ManualLocation instead of just manualLocation, ill post a pr to fix it soon

FuzzyGamesOn (approved):
I saw that you added an item value of stars to Jill and I thought it was a lead-up to an elaborate Resident Evil joke, but it wasn't. And now I'm sad.

Anyways, this looks good. 😃

silasary (approved):

PR #41, opened 2024-04-18
Ability to mark locations as always pre-hinted

silasary:
Code stolen straight from Hollow Knight's shops


nicopop (approved):
simple but it looks like it should work 👍

FuzzyGamesOn (approved):

PR #45, opened 2024-04-23
add conversion of existing goal option as alias for new system

nicopop:
~im adding this so if someone (like me) used a goal option in the options.py hook before the overhaul it will convert the values to be alias to the new system, it will also grab the \_\_doc\_\_ string that's used to explain the options
the only down side is when they convert their victory locations in locations.json they need to be in the same order as before so the int values stay the same
As a bonus existing yaml with the goal option defined will work since it will use the alias and if you do it correctly then it uses corresponding victory in the new system
also when we add a options.json if its run before this then it should also be compatible (I might work on it soonish)~
Changed to a Warning based on feedback and added code to hook

### DIFF
--- a/src/Options.py
+++ b/src/Options.py
@@ -16,12 +16,13 @@ if len(victory_names) > 1:
     goal = {'option_' + v: i for i, v in enumerate(victory_names)}
     docstring = "Choose your victory condition."
     defaultvalue = 0
-    if manual_options.get('goal'): 
-    #Grab existing Goal option and make alias out of it's values
-        for alias, value in manual_options.get('goal').options.items():
-            goal[f"alias_{alias}"] = value
-        docstring = manual_options.get('goal').__doc__
-        defaultvalue = manual_options.get('goal').default
+    # Check for existing Goal option and make alias out of it's values if possible
+    if manual_options.get('goal'):
+        if issubclass(manual_options['goal'], Choice):
+            for alias, value in manual_options['goal'].options.items():
+                goal[f"alias_{alias}"] = value
+            defaultvalue = manual_options['goal'].default
+        docstring = manual_options['goal'].__doc__ or docstring
     manual_options['goal'] = type('goal', (Choice,), goal)
     manual_options['goal'].__doc__ = docstring
     manual_options['goal'].default = defaultvalue
--- a/src/Options.py
+++ b/src/Options.py
@@ -4,6 +4,7 @@ from .hooks.Options import before_options_defined, after_options_defined
 from .Data import category_table
 from .Locations import victory_names
 from .Items import item_table
+import logging
 
 
 class FillerTrapPercent(Range):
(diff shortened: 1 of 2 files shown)


silasary (approved):

FuzzyGamesOn:
I'm confused on what this is adjusting for. It says it's converting old goal options that people might have in hooks, but I don't see where we provide anything by default for that.

Can you provide an example of the previous functionality that this would account for?

nicopop:
> I'm confused on what this is adjusting for. It says it's converting old goal options that people might have in hooks, but I don't see where we provide anything by default for that.
> 
> Can you provide an example of the previous functionality that this would account for?

in the options.py hooks it was possible to make an option of type choice called goal before this existed, of cource the logic was then only decided via hooks but before this pr the new goal would override completely the existing goal options,
what this pr does is convert all the possible values from the old goal to alias so yaml and even other hooks's logic will still work if the ids of the value match up with the new system

Now that I think about it I probably need to add a check to verify that the old goal was a choice option(ill do it later today or tomorrow
eg.
my [OuterWilds Option.py hook](https://github.com/nicopop/NicosManualApworlds/blob/ceb0ded5faf87f1b73a6f209f635127b39dd1df3/worlds/manual_outerwilds_nicopopxd/hooks/Options.py#L79-L102) 
and [OuterWilds locations.json](https://github.com/nicopop/NicosManualApworlds/blob/ceb0ded5faf87f1b73a6f209f635127b39dd1df3/worlds/manual_outerwilds_nicopopxd/data/locations.json#L754-L816)

FuzzyGamesOn:
> in the options.py hooks it was possible to make an option of type choice called goal before this existed, of cource the logic was then only decided via hooks but before this pr the new goal would override completely the existing goal options, what this pr does is convert all the possible values from the old goal to alias so yaml and even other hooks's logic will still work if the ids of the value match up with the new system

Right, this is what I'm asking about. 😃 What I'm hearing (which might be wrong, seeking clarification) is this:

- Some people defined their own options in their own hooks, sometimes called "goal"
- The logic for those options exists entirely in their hooks, and not in Manual source
- The introduction of the Victory Overhaul PR might conflict with what people defined in their own hooks

If that's the case, Manual "core" should -- as a general rule -- not try to account for whatever people are doing in hooks. We don't want to manage or vet their hooks at all... and we don't want to introduce code that changes their hooks code at all, either, beyond the occasional breaking change for official features/improvements. When we need to change something that may affect their hooks, we can notify them of the breaking change via an unstable release, which gives them time to change their hooks. But it's the responsibility of the world creator to update their hooks in those cases. 

If we want to add a warning about an existing option name and have that be loud for people to recognize a problem, that's fine to add to Manual core. If we want to rename our victory goal option to avoid a likely collision, that's okay too. The difference is that we're warning them about a problem they need to fix or trying to avoid introducing the problem in general, not trying to fix the problem for them without knowing their implementation in all cases. 😃

silasary:
Thinking about it further, this feels like it's safer to just ask the user to configure the goal with `after_options_defined`

nicopop:
I rewrote this so its now a warning and moved the conversion code to the options hooks as a commented out example

FuzzyGamesOn:
Revised changes look good, but good with a new PR if you'd rather do that 😃

nicopop:
this will be back ;)

PR #46, opened 2024-04-23
Deathlink button!

silasary:
It works!

Probably needs some kind of audio alert for when you recieve one, because button going red isn't overly attention grabbing.  But it all works.

Tested in a 2-player game with a Factorio, successfully sent and received deaths

### DIFF
--- a/schemas/Manual.game.schema.json
+++ b/schemas/Manual.game.schema.json
@@ -29,7 +29,7 @@
             }
         },
         "death_link": {
-            "description": "(Optional) Does your game have support Deathlink?",
+            "description": "(Optional) Does your game support Deathlink?",
             "type": "boolean"
         },
         "_comment": {"$ref": "#/definitions/comment"}
--- a/schemas/Manual.game.schema.json
+++ b/schemas/Manual.game.schema.json
@@ -28,6 +28,10 @@
                 "$ref": "#/definitions/Sitems"
             }
         },
+        "death_link": {
+            "description": "(Optional) Does your game have support Deathlink?",
+            "type": "boolean"
+        },
         "_comment": {"$ref": "#/definitions/comment"}
     },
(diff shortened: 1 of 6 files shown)


schemas/Manual.game.schema.json
```diff
@@ -28,6 +28,10 @@
                 "$ref": "#/definitions/Sitems"
             }
         },
+        "death_link": {
+            "description": "(Optional) Does your game have support Deathlink?",
```
nicopop, line 32: might want to reRead that line ;)

src/data/game.json
```diff
@@ -22,5 +23,6 @@
         {
             "random": 3
         }
-    ]
-}
\ No newline at end of file
+    ],
+    "death_link": true
```
nicopop, line 27: lets maybe not have it enabled by default? I feel like it should be something the creator has to choose to turn on

silasary:
My reasoning is mostly that Deathlink doesn't work for every kind of game, and presenting the yaml option for something like Solitaire would be confusing.  

I'm happy with it being default-on if you'd prefer (it was on-by-default before nico's feedback), but I think giving the world creator the choice of whether Deathlink makes sense for their world is a good idea.

FuzzyGamesOn:
> My reasoning is mostly that Deathlink doesn't work for every kind of game, and presenting the yaml option for something like Solitaire would be confusing to the end user.
> 
> I'm happy with it being default-on if you'd prefer (it was on-by-default before nico's feedback), but I think giving the world creator the choice of whether Deathlink makes sense for their world is a good idea.

That makes sense, yeah. Hard to know if people would want this available by default or gatekept by the world creator / anyone modifying the world. Without knowing, opt-in seems fine. We can always change it later based on feedback. 😃

FuzzyGamesOn (approved):

PR #48, opened 2024-04-29
Add $schema property to every json

nicopop:
I went ahead and did it today,
the important code is in data.py where I added the convert_to_list function to be backward compatible with the existing code
if "data" is not the property name we want for storing the list of items/locations then that can be changed
Since this added tabulation in the locations and items jsons this pr will conflict with every other pr that modify those files
I know #46 is already adding the $schema properties in game.json but I felt it better to have it all in one pr

### DIFF
--- a/schemas/Manual.items.schema.json
+++ b/schemas/Manual.items.schema.json
@@ -1,77 +1,89 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://raw.githubusercontent.com/ManualForArchipelago/Manual/main/schemas/Manual.items.schema.json",
-  "description": "Schema for ManualAP items.json",
-  "type": "array",
-  "items": {
-    "$ref": "#/definitions/Item"
-  },
-  "definitions": {
-    "Item": {
-      "type": "object",
-        "properties": {
-            "name": {
-                "description": "The unique name of the item. Do not use () or : in the name",
-                "type": "string"
-            },
-            "progression": {
-                "description": "(Optional) Is this item needed to unlock locations? Defaults to false if not included.",
-                "default": true,
-                "type": "boolean"
-            },
-            "count": {
-                "description": "(Optional) Number of copy of this item in the pool.",
-                "type": "integer",
-                "default": 1
-            },
-            "category": {
-                "description": "(Optional) The category(ies) this item fit in",
-                "type": "array",
-                "items": {
-                    "$ref": "#/definitions/Category"
-                },
-                "minItems": 1,
-                "uniqueItems": true
-            },
-            "useful": {
-                "description": "(Optional) Is this item usefull to have? Used for items that are not progression but you still want the rando to really use.",
-                "type": "boolean"
-            },
-            "progression_skip_balancing": {
-                "description": "(Optional) Should this item not get included in progression balance swaps. For more info check the discord",
-                "type": "boolean",
-                "default": true
-            },
-            "trap": {
-                "description": "(Optional) Is this item a trap? Something the player doesnt want to get.",
-                "type": "boolean",
-                "default": true
-            },
-            "early": {
-                "description": "(Optional) Is this item required to be placed somewhere accessible from the start (Sphere 1)",
-                "type": "boolean",
-                "default": true
-            },
-            "local": {
-                "description": "(Optional) Is this item supposed to be only in your locations(true) or anywhere(false)",
-                "type": "boolean",
-                "default": true
-            },
-            "_comment": {"$ref": "#/definitions/comment"}
-
+    "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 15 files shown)


src/Data.py
```diff
@@ -22,12 +22,21 @@ def load_data_file(*args) -> dict:
...
+game_table = load_data_file('game.json') #dict
+item_table = convert_to_list(load_data_file('items.json'), 'data') #list
+location_table = convert_to_list(load_data_file('locations.json'), 'data') #list
+region_table = load_data_file('regions.json') #dict
+category_table = load_data_file('categories.json') or {} #dict
+meta_table = load_data_file('meta.json') or {} #dict
+
+# Removal of schemas in root of tables
+region_table.pop('$schema', '')
+category_table.pop('$schema', '')
```
silasary, line 39: This shouldn't be necessary.   
And in terms of micro-optimizations, I'm more concerned by load time than a few bytes of memory.

nicopop: its not for optimization we loop thru both as if they are real region/category elsewhere in the code

nicopop: for example in create_regions the code think $schema is a real region and try to create it.
in option.py the category get looped and call .get on $schema which is a str not a dict and thus dies

nicopop: its either a pop() here or add a check everywhere that needs it to skip if str both is valid

silasary: Ah, true. I'm happy with the pop.

nicopop:
This would work as is but the support for it will need to be added to the builder

schemas/Manual.items.schema.json
```diff
@@ -1,91 +1,103 @@
...
-                "default": 1
-            },
-            "category": {
-                "description": "(Optional) The category(ies) this item fit in",
-                "type": "array",
-                "items": {
-                    "$ref": "#/definitions/Category"
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/ManualForArchipelago/Manual/main/schemas/Manual.items.schema.json",
+    "description": "Schema for ManualAP items.json",
```
FuzzyGamesOn, line 4: Minor thing but, when we refer to Manual publicly, we want to call it either "Manual" or "Manual for Archipelago". For consistency and to avoid confusion, we should avoid variations like "ManualAP" or "manual Archipelago", so these should be updated to one of those.

schemas/Manual.schema.catalog.json
```diff
@@ -1,42 +1,42 @@
...
-        "fileMatch": [ "**/manual*/data/meta.json" ],
-        "url": "https://raw.githubusercontent.com/ManualForArchipelago/Manual/main/schemas/Manual.meta.schema.json"
-    }
-  ],
-  "version": 20240406.02
+    "$schema": "https://json.schemastore.org/schema-catalog.json",
+    "schemas": [
+        {
+            "name": "Manual items.json",
+            "description": "items.json schema file for the manual Archipelago Apworld",
```
FuzzyGamesOn, line 6: I think I missed this when it came in initially, but "schema file for <referring to Manual> Apworld" should be "schema file for <referring to Manual> APWorlds" or "schema file for APWorlds created with <referring to Manual>". (For referring to Manual, see the above comment about the two possible variations to use. 😄)

This is because "the APWorld" is too small in scope when referring to Manual in the context of it being an APWorld templating system, which can occasionally be used to create APWorld templates (like Axx's fighting game one).

FuzzyGamesOn (changes requested):
Just a couple of minor revisions needed for how we refer to Manual in text. Once those are sorted, this seems ready 😄 

I'll also make a note that the Builder will need updating to support this prior to the next release in that repo.

FuzzyGamesOn (approved):

FuzzyGamesOn:
This will be in the next unstable.

nicopop:
> This will be in the next unstable. (Builder will be broken for people using this unstable code until that unstable goes stable, because it has to support the stable code first.)

As long as they use the old data file they should still be compatible, its part of this PR to support both data files version

FuzzyGamesOn:
> > This will be in the next unstable. (Builder will be broken for people using this unstable code until that unstable goes stable, because it has to support the stable code first.)
> 
> As long as they use the old data file they should still be compatible, its part of this PR to support both data files version

Oh nice! Well, nevermind then, haha

nicopop:
> > > This will be in the next unstable. (Builder will be broken for people using this unstable code until that unstable goes stable, because it has to support the stable code first.)
> > 
> > 
> > As long as they use the old data file they should still be compatible, its part of this PR to support both data files version
> 
> Oh nice! Well, nevermind then, haha
> 
> (Once the merge conflict for this is sorted, it will get merged. 😄 )

Merge conflict fixed and I took the time to test that my backward compatibility really worked

FuzzyGamesOn (approved):

PR #49, opened 2024-04-29
Fix for "Unstable APWorld: place_item on first location causes generation failure"

FuzzyGamesOn:
Walruses don't work great in an if + and, so splitting them up.


nicopop (approved):
Looks good, I personally would've used [] or "" for false but that works

silasary (approved):

PR #50, opened 2024-05-04
Strip leading and trailing whitespace in requirement counts, and provide a better error when invalid strings are passed.

silasary:
This has happened a few times recently

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -102,14 +102,14 @@ def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
             item_base = item
             item = item.lstrip('|@$').rstrip('|')
 
-            item_parts = item.split(":")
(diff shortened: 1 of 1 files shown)


FuzzyGamesOn (approved):

src/Rules.py
```diff
@@ -102,14 +102,14 @@ def checkRequireStringForArea(state: CollectionState, area: dict):
             item_base = item
             item = item.lstrip('|@$').rstrip('|')
 
-            item_parts = item.split(":")
+            item_parts = item.split(":")  # type: list[str]
             item_name = item
```
nicopop, line 106: Personally I would add a strip() around here for when there's only 1 part, but thats not 100% required soo ill leave it to your opinion,
other then that looks good

nicopop (approved):

PR #51, opened 2024-05-04
avoid doing string comparisons when removing location buttons

silasary:
This is both faster and more resilient to errors

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -190,6 +190,7 @@ class ManualContext(SuperContext):
 
         class TreeViewButton(Button, TreeViewNode):
             victory: bool = False
+            id: int = None
 
         class TreeViewScrollView(ScrollView, TreeViewNode):
             pass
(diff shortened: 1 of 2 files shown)


FuzzyGamesOn (dismissed):

src/ManualClient.py
```diff
@@ -98,7 +98,8 @@ async def connection_closed(self):
     def suggested_game(self) -> str:
         if self.game:
             return self.game
-        return Utils.persistent_load().get("client", {}).get("last_manual_game", "Manual_{\"game\" from game.json}_{\"creator\" from game.json}")
+        from .Game import game_name  # This will at least give us the name of a manual they've installed
+        return Utils.persistent_load().get("client", {}).get("last_manual_game", game_name)
```
FuzzyGamesOn, line 102: Yeah, this will work out better than the hint text. Since more people lately have only used Builder and have been doing `Manual_{Manual_Blah_Me}_Me` due to not knowing what `game.json` is.

FuzzyGamesOn (approved):

nicopop (approved):
Ui stuff is not really my strenght but the changes you made make sense.

PR #54, opened 2024-05-23
Adding options.json

nicopop:
As asked multiple times a way to modify the options

Instead of adding the parameters everywhere its possible to create options, I regrouped those and the ability to create options in a new options.json

## why a option.json?

- Have 1 file with all the parameters of options
- for people that dont want to touch python/hooks

## What's included in options.json:

- Creating new options
- Modifiying their description
- Modifiying their display name
- Modifiying their visibility
- Grouping them in the Option Groups


silasary:
I'd prefer this be an "options" key within game.json

FuzzyGamesOn:
> I'd prefer this be an "options" key within game.json

This might be another good topic for a group discussion in the dev channel to see what everyone thinks. 

Original intent for game.json was to have meta things for the game, and have collections of standalone things go elsewhere. Which I realize means that starting items might be in the wrong place, too. 😄Reasoning behind limiting game.json to meta things would be to limit the complexity and purpose of the file, so each JSON is mostly in charge of describing one type of thing.

src/Options.py
```diff
@@ -1,21 +1,84 @@
...
+        if option_name == 'goal':
+            if option['type'] != 'Choice':
+                raise Exception("a 'goal' option must be of type 'Choice'")
+            manual_goal_override['args'] = createChoiceOptions({}, option.get('aliases', {}))
+            manual_goal_override['args']['default'] = args['default'] = option.get('default', 0)
+            manual_goal_override['args']['display_name'] = option.get('display_name', option_name)
+            manual_goal_override['description'] = convertToLongString(option.get('description', ''))
+        continue
+
+    option_type = Toggle # ! I think there might be a better way to convert option['type'] -> type but I cant find it right now
```
nicopop, line 41: if someone find me a better way for this ill be able to simplify the code quite a bit

nicopop: found "locate" currently testing if that work

FuzzyGamesOn:
My two main questions when looking at this PR are:

1. What does this empower the user to do, beyond creating options and adding option descriptions for their category.json yaml_options and their goal option?
2. How is this template syntax easier or more understandable than creating options in code? (The syntax actually just looks like a direct port of the code approach in hooks. Template syntax _usually_ simplifies things and makes some assumptions for the user.)

Or, to put the questions more simply: 
**Who is this made for, and how is this easier than the hooks approach?**

nicopop:
Im posting this to says I saw that answer. I just don't know the best way to phrase a defence for this pr, and I'm not ignoring this.
In the meantime, here are the good points I can think of
- allow people who really don't want to touch Python to modify/define options 
- It's to be like category.json where functionality that were requested for option would be grouped

There were more but I can't remember them right now.

silasary:
This feels very overengineered, and I think I'd rather just point people at hooks for most of what it's doing.

My preference remains a simple `dict[str, str]` of `yaml_name` to `yaml_description`, because that's all people actually want to change when they're doing non-hook options.

FuzzyGamesOn:
> * allow people who really don't want to touch Python to modify/define options

Honestly, I agree that solving this problem would provide this benefit. There is some value to making options easier to work with for people that are intimidated by code. We just have to figure out what we're making easier through a no-code approach. It could be Sila's suggestion of option name and description, it could include visibility, it could even include types, or even other things. But we shouldn't model it after AP's code approach to options... We should model it how normal users would think about it and want to use it. 

And we can always start small with a name + description and see what people ask for the most. :smile:

Also agree with Sila. For what this is doing right now and how people would use it, I feel like we'd just point them to hooks instead.

nicopop:
I'll look into simplifying/rewriting this soonish, When I made this originally, I got way too focused and It became too much. I'll scrap the weird schema that I dislike now and probably only add support for toggle and choices for now using an enum/simple if else

nicopop:
got an idea to make the option.json simpler ill combine each option version together and add a simple test to check which it is:
Toggle + DefaultOnToggle: with a default value that find the correct one
Choice + TextChoice: A bool for if you allow user input
Range + NamedRange: if the dict is present then its named otherwise its Range

nicopop:
Pretty sure its now as simple as I can make it, I even simplified Archipelago's own option types.
Gone are the weird isinstance/locate now its done using 3 possible option type via if/else: Toggle, Choice, Range (currently only Toggle is realy used in manual, choice let you override some info on the goal we generate, range currently has no use in manual proper)
some decriptions might need to be tweaked but other than that idk what else to do.

src/Options.py
```diff
@@ -1,24 +1,117 @@
...
+
+        args = {'display_name': option.get('display_name', option_name)}
+
+        if option_type == "Toggle":
+            value = option.get('default', False)
+            option_class = DefaultOnToggle if value else Toggle
+
+        elif option_type == "Choice":
+            args = {**args, **createChoiceOptions(option.get('values'), option.get('aliases', {}))}
+            option_class = TextChoice if option.get("allow_user_input", False) else Choice
```
nicopop, line 62: I'll need to rename that because it sounds like it disable/enable the option or something
allow_custom_user_input might be a better parameter name

nicopop (commented):
Things to fix I thought of while trying to sleep

src/data/options.json
FuzzyGamesOn, line 1: This looks much better than the previous iteration! Was a great idea to reduce it down to just Toggle + Choice + Range and work off of that, and this feels natural to work with.

Few thoughts so far:
- For Choice - Agree that "allow_user_input" feels like the wrong term. To be consistent with the use of "values", maybe "allow_custom_value" or something similar. So it relates back to what is being allowed.
- For Range - "values_label" should probably just be "values", for consistency with similar Choice syntax.
- The "goal" option should allow limited customization, since Manual wants to define it and use it for core things. So the user should be able to only specify description and display_name (and _comment) for that and nothing else. We should ignore props that set the option type, values, or similar for built-in options.

nicopop: I felt like letting the dev change the default goal could also be a good idea.
same for adding alias to their goal
of cource the user cant change the built in options's type, ill even remove that check, unfortunatly its a bit harder to tell the schema to not require the type when its a goal but I might be able to do it idk ill Try

FuzzyGamesOn: As long as their changes don't break what we're doing with our built-in goal option, those changes are fine. So aliases is probably fine :smiley:

Realized I didn't mention visibility earlier, too. For the specific use case of a user wanting to just outright "replace" the goal option, probably would be nice to give them a "hidden" prop that sets visibility to none. Then that is consistent with categories terms and also with our current advice of telling them to hide goal and make their own goal-like option.

nicopop: > Realized I didn't mention visibility earlier, too. For the specific use case of a user wanting to just outright "replace" the goal option, probably would be nice to give them a "hidden" prop that sets visibility to none. Then that is consistent with categories terms and also with our current advice of telling them to hide goal and make their own goal-like

so basically I add a param to the goal only override to disable our goal system?
because otherwise setting ours to hidden would just make it invisible and still work the same as if goal == default value
and then the only way to change its value is to use hooks

FuzzyGamesOn: I think you're overcomplicating the functionality to allow users to both: (1) define their own custom options and (2) override some params of our built-in options.

The option processing shouldn't happen until our built-in options are defined. Then we have a list of our built-in option names. And when we loop their user-defined options after, if the user-defined option name matches one of our built-in option names, we only allow them to update the props that we support updating. That way, we support future options without remembering to update a separate list, and we don't have to keep track of specific overrides.

---

For the visibility thing: It's mainly for the cosmetic benefit of the user. Some users like to only see their custom goal (not named "goal") option in the YAML, so we've been recommending that they use hooks to change the visibility of our goal option to make it not show up. Which, you're correct, doesn't remove it... just allows our code to continue to function, and they take whatever victory location they derive through their own methods and update our built-in option to make use of the multiple goal locations. Or they do hooks and customize it all after, whatever works. :smiley:

FuzzyGamesOn: Also, currently, the only way to change any of the built-in option values is hooks. We're not seeking to change that with this PR, since they can't even change their own custom option values without hooks either :smiley:

nicopop: the problem with doing ours first then the custom override after is that we would need recreate the option since im pretty sure you cant add  or change arguments of an existing option (I could be wrong im testing RN)

nicopop: actually it possible but difficult ill move the base options to the top then recreate the options on override (since its easier)

nicopop: b131807 posted here's what I added on top of what we talked:
- made it super easy to add a new override in the schema in the future
- same for option.py's option.json loop
- removed the manual_override dict
- moved our options above options.json options
- added the "hidden" parameter that basically do "visibility": "none"
  - I left visibility as an advanced option

FuzzyGamesOn: Oh, I didn't realize you had visibility already. In that case, no need for "hidden", visibility covers it. My bad :smile:

FuzzyGamesOn: Added a thread here for discussion to cut down on comment chains while we figure stuff out: https://discord.com/channels/1097532591650910289/1311063001469812807

nicopop:
reposting this because it's in a resolved conversation
> Added a thread here for discussion to cut down on comment chains while we figure stuff out: https://discord.com/channels/1097532591650910289/1311063001469812807

silasary (approved):

schemas/Manual.options.schema.json
```diff
@@ -51,16 +51,16 @@
             "type": "object",
             "properties": {
                 "display_name": {
-                    "description": "The name shown on the webhost.",
+                    "description": "(Optional) The name shown on the webhost",
```
silasary, line 54: More importantly, this is used by the spoiler file

silasary (changes requested):

silasary (dismissed):

src/data/options.json
```diff
@@ -0,0 +1,82 @@
...
+            ],
+            "values": {
+                "start":0,
+                "test":1
+             },
+            "default": 0,
+            "allow_custom_value": false,
+            "group": "Example Options"
+        },
+        "Range":{
```
FuzzyGamesOn, line 54: A minor thing, but since people get confused easily... It'd be better to have the option name in the examples be something other than the type. Like "My_Toggle", "My_Choice", or something. Then it's more clear to them that the dict keys are the option name and not the type of option. (Can already foresee someone asking about this, lol.)

src/Helpers.py
```diff
@@ -155,3 +155,8 @@ def get_items_with_value(world: World, multiworld: MultiWorld, value: str, playe
                             and i.name in world.item_name_groups.get(f'has_{value}_value', [])}
         world.item_values[player][value] = item_with_values
     return world.item_values[player].get(value)
+
+def convertToLongString(input: str | list) -> str: #Todo maybe find a better name for this
```
FuzzyGamesOn, line 159: Another minor thing, but function is camel case, but other functions in helpers are lower w/ underscores. So this should also be lower w/ underscores for consistency, since people might also be using the helpers in their hooks.

FuzzyGamesOn (changes requested):
The most minor of changes. Rest looks great :smiley:

FuzzyGamesOn (approved):

silasary (approved):

PR #55, opened 2024-05-24
Fix height of Game ID field using Density Independent Pixels

kedNalatacId:
Use the same method as the base kvui library to properly scale the Game ID field's height based on pixel density.

Only tested on Mac (where it was an issue).

Old behavior:
Screenshot 2024-05-24 at 12 12 08

New behavior:
Screenshot 2024-05-24 at 12 12 22


FuzzyGamesOn (approved):
This looks good, yep. Thanks for the contribution! 😄

PR #57, opened 2024-06-10
Fix multiworld.get_items() doesnt include precollected Items & ItemValue

nicopop:
Original Title: Fix ItemValue cache breaks if used multiple times in a single requirement
~~based on this issue:
[on discord](https://discord.com/channels/1097532591650910289/1249649852372095048)
Seems I forgot to test ItemValue when its called for multiple time in the same requires and/or with multiple value types
While I made this I discovered that multiworld.get_items() does NOT return precollected items thus breaking ItemValue even more
this probably also break item_counts. (fix included)~~
# TLDR
while fixing [this issue on discord](https://discord.com/channels/1097532591650910289/1249649852372095048) I discovered that precollected items are not returned in multiworld.get_items() when ItemValue's validation sometime failed using the Manual_UltimateMarvelVsCapcom3_ManualTeam starting_items.
I realised this also affected item_counts.
oh, Also when I made ItemValue originally I forgot to test using multiple value types at the same time... but that became the secondary objective of this pr...
# What This PR Actually changes
- Make the requires function detection allow `)` in the arguments using a non-greedy/lazy regex
  - Will only break if exactly `)}` is in the function arguments
- Added the `includePrecollected: bool` param to get_items_for_player
  - With this enabled get_items_for_player include precollected items in the list
  -  Before this there was a logic bug with using starting_items and anything that uses the item_counts
- Made ItemValue actually work with multiple Value types
- Changed ItemValue's validation regex to also not be greedy/lazy
- Added a `skipCache: bool` Param for ItemValue

### DIFF
--- a/src/hooks/Rules.py
+++ b/src/hooks/Rules.py
@@ -28,39 +28,47 @@ def requiresMelee(world: World, multiworld: MultiWorld, state: CollectionState, 
     """Returns a requires string that checks if the player has unlocked the tank."""
     return "|Figher Level:15| or |Black Belt Level:15| or |Thief Level:15|"
 
-def ItemValue(world: World, multiworld: MultiWorld, state: CollectionState, player: int, args: str):
+def ItemValue(world: World, multiworld: MultiWorld, state: CollectionState, player: int, valueCount: str, noCache: Optional[str] = None):
     """When passed a string with this format: 'valueName:int',
     this function will check if the player has collect at least 'int' valueName worth of items\n
-    eg. {ItemValue(Coins:12)} will check if the player has collect at least 12 coins worth of items
+    eg. {ItemValue(Coins:12)} will check if the player has collect at least 12 coins worth of items\n
+    You can add a second string argument to disable creating/checking the cache like this:
+    '{ItemValue(Coins:12,Disable)}' it can be any string you want
     """
 
-    args_list = args.split(":")
-    if not len(args_list) == 2 or not args_list[1].isnumeric():
-        raise Exception(f"ItemValue needs a number after : so it looks something like 'ItemValue({args_list[0]}:12)'")
-    args_list[0] = args_list[0].lower().strip()
-    args_list[1] = int(args_list[1].strip())
+    valueCount = valueCount.split(":")
+    if not len(valueCount) == 2 or not valueCount[1].isnumeric():
+        raise Exception(f"ItemValue needs a number after : so it looks something like 'ItemValue({valueCount[0]}:12)'")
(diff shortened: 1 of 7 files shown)


src/hooks/Rules.py
```diff
@@ -28,39 +28,47 @@ def requiresMelee(world: World, multiworld: MultiWorld, state: CollectionState,
...
     return "|Figher Level:15| or |Black Belt Level:15| or |Thief Level:15|"
 
-def ItemValue(world: World, multiworld: MultiWorld, state: CollectionState, player: int, args: str):
+def ItemValue(world: World, multiworld: MultiWorld, state: CollectionState, player: int, valueCount: str, noCache: Optional[str] = None):
     """When passed a string with this format: 'valueName:int',
     this function will check if the player has collect at least 'int' valueName worth of items\n
-    eg. {ItemValue(Coins:12)} will check if the player has collect at least 12 coins worth of items
+    eg. {ItemValue(Coins:12)} will check if the player has collect at least 12 coins worth of items\n
+    You can add a second string argument to disable creating/checking the cache like this:
+    '{ItemValue(Coins:12,Disable)}' it can be any string you want
```
FuzzyGamesOn, line 36: The two places that use `noCache` below use it as a bool. So we'd either want it to be bool in the type hint above (for true/false) or int (for 0/1). Whichever seems most convenient for people building their worlds. 😄 

The way it is now, I could put "enabled" or "on" or "always_cache", and all of those would disable the cache, haha.

FuzzyGamesOn (changes requested):

src/hooks/Rules.py
FuzzyGamesOn: That makes sense, yeah. I think the goal should be to remove ambiguity as much as possible. Non-empty string always evaluating to True is less intuitive than either (1) converting string True/False/true/false in func args to bool or (2) using 0 and 1 and doing a `bool()` call. Just unfortunately something that wasn't considered with req functions because every arg is a string by default.

nicopop: I could write some code to identify the type of a require fonction args and then try and convert the user's input to said type and probably raise exception if the conversion I program fail? that might be something for another pr

nicopop: I got some argument conversion code ready I just dont know if I add it to this pr or if we make another

silasary (approved):

FuzzyGamesOn (approved):
Awesome, this is good to go! Noting that this has to merge with https://github.com/ManualForArchipelago/Manual/pull/65 or the skipCache arg will not work as expected.

PR #58, opened 2024-06-14
Client: Hinted locations without a category key will error

FuzzyGamesOn:
This was noticed with a very old apworld that likely predated the addition of categories to Manual. The part of the client that displays hinted locations doesn't check for the "category" key to be on the location before checking it / appending to it.

So this PR sets the location "category" to an empty list if not present.


axxroytovu (approved):
Seems clean enough

silasary (approved):

PR #59, opened 2024-06-16
{YamlEnabled()} function

silasary:
The PR does a couple things:

- Added YamlEnabled/Disabled functions
- ~~Moved Predefined functions into non-hook Rules.py~~
- Added ability to load predefined functions from non-hook Rules.py
- Added documentation for Predefined functions

The new functions should be obvious (If they're not, hopefully the new documentation helps).

The refactor is to make things less messy when people update their apworlds.  It felt weird to have stuff living in hooks that isn't user code (Especially once I started writing documentation for it)

PR #60, opened 2024-06-20
Add canReachLocation

axxroytovu:
Adds a new canReachLocation available rule, to expose the core `can_reach_location` function of the CollectionState. This is a commonly requested feature and could stand-in for events for logic building without requiring a bunch of extra clicks.


nicopop (approved):
seems good

silasary (approved):

axxroytovu:
If we do move the "non hook" functions to a different location re: https://github.com/ManualForArchipelago/Manual/issues/59, then this function should get moved as well.

FuzzyGamesOn:
The hook functions move PR (#59) will be in the next unstable, and this will get moved then. 😃

PR #61, opened 2024-06-24
Don't crash if a meta.json isn't provided

silasary:
If Meta.json is missing or empty, set_world_description throws a KeyError, and prevents the world from loading.

This makes it fail gracefully.


nicopop (approved):
Looks good

PR #62, opened 2024-07-02
Add local_early option and allow some copy of an item to be early

nicopop:
A month or so ago I programed this and forgot to do anything with it.
So here's a way to only mark some copies of an item as early and/or local_early
unfortunatly 'local' is still a none or all option but thats a AP thing
probably a good idea to wait until #48 merge so I can properly add my modification to it

### DIFF
--- a/schemas/Manual.items.schema.json
+++ b/schemas/Manual.items.schema.json
@@ -1,91 +1,96 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://raw.githubusercontent.com/ManualForArchipelago/Manual/main/schemas/Manual.items.schema.json",
-  "description": "Schema for ManualAP items.json",
-  "type": "array",
-  "items": {
-    "$ref": "#/definitions/Item"
-  },
-  "definitions": {
-    "Item": {
-      "type": "object",
-        "properties": {
-            "name": {
-                "description": "The unique name of the item. Do not use () or : in the name",
-                "type": "string"
-            },
-            "progression": {
-                "description": "(Optional) Is this item needed to unlock locations? Defaults to false if not included.",
-                "default": true,
-                "type": "boolean"
-            },
-            "count": {
-                "description": "(Optional) Number of copy of this item in the pool.",
-                "type": "integer",
-                "default": 1
-            },
-            "category": {
-                "description": "(Optional) The category(ies) this item fit in",
-                "type": "array",
-                "items": {
-                    "$ref": "#/definitions/Category"
+    "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -117,16 +117,35 @@ def create_items(self):
...
+            for _ in range(item_count):
                 new_item = self.create_item(name)
                 pool.append(new_item)
 
-            if item.get("early"): # only early
-                self.multiworld.early_items[self.player][name] = item_count
-            if item.get("local"): # only local
+            if item.get("early"): # Some or all early
+                if isinstance(item["early"],int) or (isinstance(item["early"],str) and item["early"].isnumeric()):
+                    self.multiworld.early_items[self.player][name] = int(item.get("early"))
```
axxroytovu, line 126: Seems redundant to check if early is an int then use `get` instead of just referencing the item directly. Is there something I'm missing?

axxroytovu (dismissed):

axxroytovu:
One minor comment but otherwise looks good. I don't have the setup to test the json schemas so someone else will need to validate those.

src/__init__.py
nicopop: nope thats a mistake, Im leaving the int() so it convert if its a valid string but the `get` is a leftover from last version of this code

axxroytovu (dismissed):

FuzzyGamesOn (dismissed):

nicopop:
this commit will conflict with #48 so either we need to merge that one first then ill patch this one or the reverse

FuzzyGamesOn (approved):

schemas/Manual.items.schema.json
axxroytovu, line 1: I think the `default` values are wrong across the board. 

The default should be "if I don't include this parameter, what is the behavior". With that interpretation, all items are progression and traps, all items are local, and one copy of every item is early.

axxroytovu (changes requested):

schemas/Manual.items.schema.json
nicopop: The way default work with vscode at least is that when you auto complete the parameters it write the default value, I know it's not exactly how it shoupd be but I made it like that because its more helpful that way than auto writing false each time and asking the dev to change it

axxroytovu: Ngl, I really dislike that but I get it.

axxroytovu (approved):

schemas/Manual.items.schema.json
nicopop: just checked a "default" value for a schema is quote "A default value. Used by suggestions." its not necessarily linked to logic or anything it can just there for auto complete/to suggest a value

FuzzyGamesOn: Noting that we discussed in the dev channel and adjusted default values are coming in #68 😄

PR #63, opened 2024-07-04
Fix the `get_filler_item_name` function

axxroytovu:
Any "item link" or "starting item from pool" references the base world function `get_filler_item_name`

By default the `get_filler_item_name` function will return a random item from the multiworld, which could break logic by adding extra progression items.

This changes that to return the normal `filler_item_name` by default, with a hook to allow the user to change the list to be anything that makes sense for that game.

### DIFF
--- a/src/hooks/World.py
+++ b/src/hooks/World.py
@@ -115,10 +115,6 @@ def before_create_item(item_name: str, world: World, multiworld: MultiWorld, pla
 def after_create_item(item: ManualItem, world: World, multiworld: MultiWorld, player: int) -> ManualItem:
     return item
 
-# Use this function to change the valid filler items to be created to replace item links or starting items
-def get_filler_item_name() -> str:
-    return filler_item_name
-
 # This method is run towards the end of pre-generation, before the place_item options have been handled and before AP generation occurs
 def before_generate_basic(world: World, multiworld: MultiWorld, player: int) -> list:
     pass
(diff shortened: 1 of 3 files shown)


silasary (approved):
Looks good

src/hooks/World.py
```diff
@@ -114,6 +115,10 @@ def before_create_item(item_name: str, world: World, multiworld: MultiWorld, pla
 def after_create_item(item: ManualItem, world: World, multiworld: MultiWorld, player: int) -> ManualItem:
     return item
 
+# Use this function to change the valid filler items to be created to replace item links or starting items
+def get_filler_item_name() -> str:
+    return filler_item_name
+
```
FuzzyGamesOn, line 121: I think it would make more sense for this to have a default implementation in the Manual world itself, then the hook would allow overriding it like other hooks do. So we don't have core required functionality living in hooks files.

Could be a before or after, or could be a "hook" like interpret_slot_data uses.

FuzzyGamesOn (changes requested):

src/hooks/World.py
axxroytovu: Ok, reworking to behave like `interpret_slot_data`

Does it make sense to leave the hook in the `World.py` file, or does it belong in the `Items.py` hook file?

FuzzyGamesOn: Sorry, just saw this. World.py makes more sense since it's a method on the world object, like create_item, etc. Basically, each of the hook files should provide a hook for the same named file that the method came from, so Item/Location hook files are basically just "before or after item/location table processing".

FuzzyGamesOn: Forgot to ping on above late comment @axxroytovu 😃

silasary (dismissed):
Looks good to me

src/hooks/World.py
axxroytovu: Fixed

FuzzyGamesOn (approved):

silasary (approved):

PR #64, opened 2024-07-04
Small Deprecation fix

nicopop:
its a tiny pr that just replace 
`multiworld.start_location_hints[player].value.add(location)`
with
`world.options.start_location_hints.value.add(location)`
since accessing start_location_hints through multiworld is now deprecated in 0.5
it still gen without but it gives a warning sooo here's my tiny pr


silasary (approved):

FuzzyGamesOn (approved):
Good catch! 😁

PR #65, opened 2024-07-08
Convert Req functions args to real type

nicopop:
has proposed [here](https://github.com/ManualForArchipelago/Manual/pull/57#discussion_r1667622795)

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -8,6 +8,7 @@ from worlds.AutoWorld import World
 
 import re
 import math
+import inspect
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


nicopop:
currently break if TypingOptionnal like OptOne, working on a fix

FuzzyGamesOn (approved):
Looks great! 😁

silasary (approved):
I'm happy enough with this.  

Technically, it's a broken implementation, as it can't handle `typing.List` and friends, but right now we don't support those anyway, so I'm happy to ship it and fix that if it comes up.

PR #68, opened 2024-07-12
Schema adjustments for defaults, structure for Builder use, etc.

FuzzyGamesOn:
We were talking about default values in schema and I had some in-progress changes to schema as well, so figured I might as well do the defaults bit and get all the changes in. 😄 

The new version of the Manual Builder (WIP) heavily uses the schema files to know what the structure of objects in the JSON files should be. This will enable us to make regular updates to Manual and its structure without having to remember (in most cases) to make updates to Builder to support it as well.

Things I needed to change about the schema to support this:

- The ordering of items needed to change to an order of relative importance, so fields were moved around
- Simple collections like arrays of strings sometimes had definition entries instead of just being items: type string, and others had items: type string, so this was made consistent to the latter.
- Some typos in schema, like categories having definitions for requires/connects_to.
- Default values that represent what the value would be if the user doesn't provide one. This is helpful to Builder because it lets me display defaults if there are any, letting the user then edit them.

I think I got all the defaults, but let me know if I missed anything or mucked up anything. Not as familiar with schema things as Nico 😄

### DIFF
--- a/schemas/Manual.items.schema.json
+++ b/schemas/Manual.items.schema.json
@@ -26,48 +26,19 @@
                     "description": "The unique name of the item. Do not use () or : in the name",
                     "type": "string"
                 },
-                "progression": {
-                    "description": "(Optional) Is this item needed to unlock locations? Defaults to false if not included.",
-                    "default": true,
-                    "type": "boolean"
+                "category": {
+                    "description": "(Optional) The category(ies) this item fit in",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "minItems": 1,
+                    "uniqueItems": true
                 },
                 "count": {
                     "description": "(Optional) Number of copy of this item in the pool.",
                     "type": "integer",
                     "default": 1
-                },
-                "category": {
-                    "description": "(Optional) The category(ies) this item fit in",
-                    "type": "array",
-                    "items": {
-                        "$ref": "#/definitions/Category"
-                    },
-                    "minItems": 1,
-                    "uniqueItems": true
-                },
-                "useful": {
-                    "description": "(Optional) Is this item usefull to have? Used for items that are not progression but you still want the rando to really use.",
-                    "type": "boolean"
-                },
-                "progression_skip_balancing": {
-                    "description": "(Optional) Should this item not get included in progression balance swaps. For more info check the discord",
-                    "type": "boolean",
-                    "default": true
-                },
-                "trap": {
-                    "description": "(Optional) Is this item a trap? Something the player doesnt want to get.",
-                    "type": "boolean",
-                    "default": true
-                },
-                "early": {
-                    "description": "(Optional) Is this item required to be placed somewhere accessible from the start (Sphere 1)",
-                    "type": "boolean",
-                    "default": true
-                },
-                "local": {
-                    "description": "(Optional) Is this item supposed to be only in your locations(true) or anywhere(false)",
-                    "type": "boolean",
-                    "default": true
                 },
                 "value": {
                     "description": "(Optional) A dictionary of values this item has in the format {\"name\": int,\"otherName\": int} \nUsed with the {ItemValue(Name: int)} in location requires \neg. \"value\": {\"coins\":10} mean this item is worth 10 coins",
@@ -83,13 +54,44 @@
                         }
                     }
                 },
+                "progression": {
+                    "description": "(Optional) Is this item needed to unlock locations? Defaults to false if not included.",
+                    "type": "boolean",
+                    "default": false
+                },
+                "progression_skip_balancing": {
+                    "description": "(Optional) Should this item not get included in progression balance swaps. For more info check the discord",
+                    "type": "boolean",
+                    "default": false
+                },
+                "useful": {
+                    "description": "(Optional) Is this item usefull to have? Used for items that are not progression but you still want the rando to really use.",
+                    "type": "boolean",
+                    "default": false
+                },
+                "trap": {
+                    "description": "(Optional) Is this item a trap? Something the player doesnt want to get.",
+                    "type": "boolean",
+                    "default": false
+                },
+                "filler": {
+                    "description": "(Optional) Is this item a filler item? Something that is mostly useless to the player.",
+                    "type": "boolean",
+                    "default": true
+                },
+                "early": {
+                    "description": "(Optional) Is this item required to be placed somewhere accessible from the start (Sphere 1)",
+                    "type": "boolean",
+                    "default": false
+                },
+                "local": {
+                    "description": "(Optional) Is this item supposed to be only in your locations(true) or anywhere(false)",
+                    "type": "boolean",
+                    "default": false
+                },
                 "_comment": {"$ref": "#/definitions/comment"}
-
             },
             "required": ["name"]
-        },
-        "Category": {
-            "type": "string"
         },
         "comment": {
             "description": "(Optional) Does nothing, Its mainly here for Dev notes for future devs to understand your logic",
--- a/schemas/Manual.items.schema.json
+++ b/schemas/Manual.items.schema.json
@@ -27,7 +27,7 @@
                     "type": "string"
                 },
                 "category": {
-                    "description": "(Optional) The category(ies) this item fit in",
+                    "description": "(Optional) A list of categories to be applied to this item.",
                     "type": "array",
                     "items": {
                         "type": "string"
@@ -36,7 +36,7 @@
                     "uniqueItems": true
                 },
                 "count": {
-                    "description": "(Optional) Number of copy of this item in the pool.",
+                    "description": "(Optional) Total number of this item that will be in the itempool for randomization.",
                     "type": "integer",
                     "default": 1
                 },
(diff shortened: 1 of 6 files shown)


nicopop (approved):
looks good.

schemas/Manual.items.schema.json
```diff
@@ -26,76 +26,78 @@
...
+                                {
+                                    "type": "integer",
+                                    "description": "A value that this item has must be 'name':integer \neg. \"coins\": 10"
+                                }
+                            ]
+                        }
+                    }
+                },
+                "progression": {
+                    "description": "(Optional) Is this item needed to unlock locations? Defaults to false if not included.",
```
axxroytovu, line 58: Do we need to keep the default in the description block if we fixed how the `default` tag works?

schemas/Manual.items.schema.json
```diff
@@ -26,76 +26,78 @@
...
+                    "default": false
                 },
                 "progression_skip_balancing": {
                     "description": "(Optional) Should this item not get included in progression balance swaps. For more info check the discord",
                     "type": "boolean",
-                    "default": true
+                    "default": false
+                },
+                "useful": {
+                    "description": "(Optional) Is this item usefull to have? Used for items that are not progression but you still want the rando to really use.",
```
axxroytovu, line 68: check spelling on "useful"

schemas/Manual.locations.schema.json
```diff
@@ -46,56 +42,59 @@
                     },
                     "uniqueItems": true
                 },
-                "place_item_category": {
-                    "description": "(Optional) Specify the category of item that will be placed at this location. (Ignore Logic)",
+                "region": {
+                    "description": "(Optional) The name of the Region this location is part of.",
```
axxroytovu, line 46: Capitalization question. Should "Region" "Location" and "Item" be capitalized everywhere or nowhere? Inconsistent within this descriptor.

schemas/Manual.locations.schema.json
```diff
@@ -46,56 +42,59 @@
...
-                        "$ref": "#/definitions/Place_item"
+                        "type": "string"
                     },
                     "minItems": 0,
                     "uniqueItems": true
                 },
-                "dont_place_item_category": {
-                    "description": "(Optional) Specify the categories of items that will never be placed at this location.",
+                "place_item_category": {
+                    "description": "(Optional) Specify the category of item that will be placed at this location. (Ignore Logic)",
```
axxroytovu, line 68: Does this ignore logic? Eventually the multiworld will still check if the game is beatable.

Also we should probably include "if multiple categories are given, the chosen item is from ANY category, not ALL categories"

schemas/Manual.locations.schema.json
```diff
@@ -46,56 +42,59 @@
...
                     "uniqueItems": true
                 },
-                "place_item_category": {
-                    "description": "(Optional) Specify the category of item that will be placed at this location. (Ignore Logic)",
+                "region": {
+                    "description": "(Optional) The name of the Region this location is part of.",
+                    "type": "string"
+                },
+                "place_item": {
+                    "description": "(Optional) Specify the list of item that can be placed at this location (will only randomly choose 1).",
```
axxroytovu, line 50: Should be "list of items" not "list of item".

Also "will choose 1 at random" sounds better than "will only randomly choose 1"

schemas/Manual.regions.schema.json
```diff
@@ -29,14 +29,15 @@
...
                             "type": "array",
                             "items": {
-                                "$ref": "#/definitions/Connects_to"
+                                "type": "string"
                             },
                             "minItems": 0,
                             "uniqueItems": true
                         },
                         "starting": {
                             "description": "(Optional) Is this region accessible from the start? Defaults to false if not included.",
```
axxroytovu, line 38: Again, do we need the default in the description here?

schemas/Manual.locations.schema.json
```diff
@@ -46,56 +42,59 @@
...
                     "type": "array",
                     "items": {
-                        "$ref": "#/definitions/Place_item"
+                        "type": "string"
                     },
                     "minItems": 0,
                     "uniqueItems": true
                 },
                 "victory": {
                     "description": "(Optional) Is this location the goal of this Manual Apworld",
```
axxroytovu, line 86: Should we change this to "one of the goal(s)" since now multiple goals are supported?

schemas/Manual.items.schema.json
FuzzyGamesOn: Good catch! Removed those 😄

schemas/Manual.items.schema.json
FuzzyGamesOn: Fixed that as well. Reworked all of the item classification descriptions to be more uniform and more concisely describe themselves.

schemas/Manual.locations.schema.json
FuzzyGamesOn: I could see equal reasons to have it capitalized and not capitalized. I went with not capitalized for now since it was the most used, but we can always change it whenever. Agree that consistency is the goal either way, so made them consistent.

schemas/Manual.locations.schema.json
FuzzyGamesOn: Yeah, this is more warning the world builder that their place item will succeed even if it would cause an unreachable location during generation. i.e., it doesn't check to see that the place item itself would prevent reaching the location. Seems like a good warning to have, so preserved that with a minor adjustment.

Also, agree, tweaked the wording on items and item categories to more clearly indicate that 1 item is being picked from the matching criteria.

schemas/Manual.locations.schema.json
FuzzyGamesOn: Tweaked this as well.

schemas/Manual.regions.schema.json
FuzzyGamesOn: All defaults should be gone from the descriptions 😄

schemas/Manual.locations.schema.json
FuzzyGamesOn: Yep, good catch! Changed that to "one of the possible goals".

schemas/Manual.items.schema.json
```diff
@@ -26,76 +26,78 @@
...
                 },
                 "local": {
                     "description": "(Optional) Are all copies of this item supposed to be only in your locations(true) or anywhere(false)",
                     "type": "boolean",
-                    "default": true
+                    "default": false
                 },
                 "local_early": {
-                    "description": "(Optional) How many copies of this item are supposed to be early and only in your locations. \nTrue mark all of them to be local_early. \nCan be used to mark some of the copies of an item to be early and local since 'local' is a toggle between none or all of them.",
+                    "description": "(Optional) How many copies of this item are supposed to be early and only in your locations. \nCan be used to mark some of the copies of an item to be early and local since 'local' is a toggle between none or all of them.",
```
nicopop, line 93: marking this as true is a way to mark all copies as local_early I do agree it could've been in a better order something like
```suggestion
                    "description": "(Optional) How many copies of this item are supposed to be early and only in your locations. \nCan be used to mark some of the copies of an item to be early and local since 'local' is a toggle between none or all of them. \nYou can also set this to 'true'  to have every copy of this item be in local_early",
```

schemas/Manual.locations.schema.json
```diff
@@ -26,76 +26,75 @@
...
-                        "$ref": "#/definitions/Place_item_category"
+                        "type": "string"
                     },
                     "minItems": 0,
                     "uniqueItems": true
                 },
-                "place_item": {
-                    "description": "(Optional) Specify the list of item that can be placed at this location (will only randomly choose 1).",
+                "dont_place_item": {
+                    "description": "(Optional) Places an item that DOES NOT match one of the item names listed in this setting at this location. Does not check logical access to the location.",
```
nicopop, line 59: `dont_place_item_*` work in 2 ways, 
1. when  combined with `place_item_*` it remove from the placement list and thus ignore logic 
2. when alone, all it does is call forbid and does not force an item out of logic there thus it does check access if I understand how AP work

its not `dont_place_item_*` that force place an item basicaly, its `place_item_*`
All they do is blacklist items from beeing placed there, the way its done just depend if place_item is present or not

schemas/Manual.locations.schema.json
```diff
@@ -26,76 +26,75 @@
...
-                        "$ref": "#/definitions/Place_item_category"
+                        "type": "string"
                     },
                     "minItems": 0,
                     "uniqueItems": true
                 },
-                "dont_place_item": {
-                    "description": "(Optional) Specify the list of items that cannot be placed at this location.",
+                "dont_place_item_category": {
+                    "description": "(Optional) Places an item that DOES NOT match at least one of the categories listed in this setting at this location. Does not check logical access to the location.",
```
nicopop, line 77: same as `dont_place_item`

nicopop (changes requested):
there are some definitions changes I dont agree with.
It seems you misunderstood how I coded dont_place_item_*

schemas/Manual.items.schema.json
FuzzyGamesOn: Made a similar change that puts the 'true' alternative value closer to the numeric suggested value.

schemas/Manual.locations.schema.json
FuzzyGamesOn: Oh no, this was just a lazy copy-paste. Wasn't intending to fix a lot of pre-existing poor wording on schema descriptions when reorganizing the files and supplying defaults, but here we are. 😄 

I've clarified both of these to indicate that it configures what item criteria should not end up at the location during normal gen. Let me know if you prefer a different explanation!

nicopop (approved):

axxroytovu (approved):

PR #70, opened 2024-07-20
Add Yaml Description

axxroytovu:
Adds the optional `"yaml_description"` item to the `Categories.json` file, which allows the user to add a line of description text to the yaml options.

Example syntax:
```json
{
    "Example Yaml-option category": {
        "yaml_option": ["DLC_enabled"],
        "yaml_description": {"DLC_enabled": "Include items and locations that require purchasing the DLC"}
    }
}
```

Result in the player yaml:
```yaml
  DLC_enabled:
    # Include items and locations that require purchasing the DLC
    'false': 0
    'true': 50
```

### DIFF
--- a/schemas/Manual.categories.schema.json
+++ b/schemas/Manual.categories.schema.json
@@ -30,6 +30,10 @@
                             "minItems": 1,
                             "uniqueItems": true
                         },
+                        "yaml_description": {
+                            "description": "(Optional) Dictionary of option-description pairs to add descriptor text to the yaml options defined in 'yaml_option'",
+                            "type": "object"
+                        }
                         "_comment": {"$ref": "#/definitions/comment"}
                     }
                 }
--- a/src/Options.py
+++ b/src/Options.py
@@ -30,6 +30,8 @@ for category in category_table:
         if option_name not in manual_options:
             manual_options[option_name] = type(option_name, (DefaultOnToggle,), {"default": True})
             manual_options[option_name].__doc__ = "Should items/locations linked to this option be enabled?"
+        if option_name in category_table[category].get("yaml_description", []):
+            manual_options[option_name].__doc__ = category_table[category]["yaml_description"][option_name]
 
 manual_options = after_options_defined(manual_options)
 manual_options_data = make_dataclass('ManualOptionsClass', manual_options.items(), bases=(PerGameCommonOptions,))
--- a/src/data/categories.json
+++ b/src/data/categories.json
@@ -4,6 +4,7 @@
         "hidden": true
     },
     "Example Yaml-option category": {
-        "yaml_option": ["DLC_enabled"]
+        "yaml_option": ["DLC_enabled"],
+        "yaml_description": {"DLC_enabled": "Include items and locations that require purchasing the DLC"}
     }
 }


schemas/Manual.categories.schema.json
```diff
@@ -30,6 +30,10 @@
                             "minItems": 1,
                             "uniqueItems": true
                         },
+                        "yaml_description": {
+                            "description": "(Optional) Dictionary of option-description pairs to add descriptor text to the yaml options defined in 'yaml_option'",
+                            "type": "object"
+                        }
```
nicopop, line 36: personally I would define the object using paternproperties but this should work as is

nicopop (approved):
this feels more like something that should be done in a dedicated option section/file like the wip #54
In the short term, yes this works but im not 100% happy with it but I leave the decisions to fuzzy

schemas/Manual.categories.schema.json
axxroytovu: Yeah I've never used schemas so I did the bare minimum here. Feel free to suggest whatever makes sense to change here.

How do patternproperties work?

axxroytovu:
> this feels more like something that should be done in a dedicated option section/file like the wip #54 In the short term, yes this works but im not 100% happy with it but I leave the decisions to fuzzy

I didn't know how far along #54 was so I wanted to push this fairly simple change in the short term.

schemas/Manual.categories.schema.json
nicopop: Basicaly normaly with an object you can define its properties using `properties` but, when they are define by the user you can use a regex aka a pattern using `patternProperties` 
you could look at [items.schema.json](https://github.com/ManualForArchipelago/Manual/blob/main/schemas/Manual.items.schema.json#L77-L90) for a pretty good example for your situation

silasary:
I do not like this syntax.

Putting a dict in each category is confusing, and the moment people start using the same option for multiple categories, they're going to be confused whether they need to duplicate the description.

If we're going to do this as a stopgap for #54, I'd want the dict in game.json, not in every category.

FuzzyGamesOn:
> I do not like this syntax.
> 
> Putting a dict in each category is confusing, and the moment people start using the same option for multiple categories, they're going to be confused whether they need to duplicate the description.
> 
> If we're going to do this as a stopgap for #54, I'd want the dict in game.json, not in every category.

I agree with sila. Wherever it is, it has to be a single place to change it. And agree with both that the best single place would be the Options json when it happens.

game.json is a good fallback, but we know we'd be building something that we're going to deprecate probably this year. 

If we're really insistent on trying to get YAML option descriptions into world builders' hands, I'd almost prefer that we just build out a skeleton Options json that just takes an option name and lets you specify an option description, and nothing else. Then Nico's WIP PR can add types and other stuff later. That way, the YAML option description ends up in the file it's going to be in. We can explain away that Options in general are a WIP and that this would be a precursor to more functionality. (And, obviously, the format for this skeleton Options json should follow what Nico is establishing in their WIP PR).

axxroytovu:
All good, I'll wait until Options.json is ready then.

PR #73, opened 2024-08-15
Add a way to decide the initial item/location id offset 

nicopop:
A simple way to set the id offset
when not specified the id offset is decided using the current system.
As of writing this PR its done by adding `"starting_index": int` in game.json but that could be changed

we might need #72 to be merged before this one so the client work with id being the same between multiple manual
idk im not a client side dev I just did what the deprecation notice message said to do.

### DIFF
--- a/schemas/Manual.game.schema.json
+++ b/schemas/Manual.game.schema.json
@@ -38,8 +38,9 @@
             "default": false
         },
         "starting_index": {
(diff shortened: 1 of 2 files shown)


silasary (dismissed):

FuzzyGamesOn (approved):
Clean 😄

silasary (approved):

PR #75, opened 2024-09-19
Allow items to skip IDs

silasary:
This is useful both for providing buffer space for future expansion (without shifting later IDs), and also just nice for organization reasons.

It's also an alternative to #73

### DIFF
--- a/src/Items.py
+++ b/src/Items.py
@@ -28,11 +28,11 @@ for key, val in enumerate(item_table):
     if "id" in item_table[key]:
         item_id = item_table[key]["id"]
         if item_id > count:
-            item_table[key]["id"] = item_id
+            count = item_id
         else:
             raise ValueError(f"{item_table[key]['name']} has an invalid ID. ID must be at least {count + 1}")
-    else:
-        item_table[key]["id"] = count
+
+    item_table[key]["id"] = count
     item_table[key]["progression"] = val["progression"] if "progression" in val else False
     count += 1
 
--- a/src/Items.py
+++ b/src/Items.py
@@ -25,7 +25,14 @@ if filler_item_name:
 
 # add sequential generated ids to the lists
 for key, val in enumerate(item_table):
-    item_table[key]["id"] = count
+    if "id" in item_table[key]:
+        item_id = item_table[key]["id"]
+        if item_id > count:
+            item_table[key]["id"] = item_id
+        else:
+            raise ValueError(f"{item_table[key]['name']} has an invalid ID. ID must be at least {count + 1}")
+    else:
+        item_table[key]["id"] = count
     item_table[key]["progression"] = val["progression"] if "progression" in val else False
     count += 1
 
--- a/src/Items.py
+++ b/src/Items.py
@@ -27,7 +27,7 @@ if filler_item_name:
 for key, val in enumerate(item_table):
     if "id" in item_table[key]:
         item_id = item_table[key]["id"]
-        if item_id > count:
+        if item_id >= count:
             count = item_id
         else:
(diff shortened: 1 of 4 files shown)


axxroytovu:
Can you add a syntax example to the `Items.json` file?

nicopop:
1. this probably need to be added to the json schemas
2. might cause id issue down the line if an id is specified and later the automated count get counted up to it

silasary:
> 1. this probably need to be added to the json schemas

Yup, can do.

> 2. might cause id issue down the line if an id is specified and later the automated count get counted up to it

This is why I throw an error if that happens.

src/Items.py
```diff
@@ -25,7 +25,14 @@
...
 for key, val in enumerate(item_table):
-    item_table[key]["id"] = count
+    if "id" in item_table[key]:
+        item_id = item_table[key]["id"]
+        if item_id > count:
+            item_table[key]["id"] = item_id
+        else:
+            raise ValueError(f"{item_table[key]['name']} has an invalid ID. ID must be at least {count + 1}")
+    else:
+        item_table[key]["id"] = count
```
FuzzyGamesOn, line 35: I think this breaks in a situation where world creators only set some of the item ids. Like, if I have these items:

- Item 1
- Item 2, id = 4
- Item 3
- Item 4

Then id 4 will be set for 2 items.

We might want to set count to the custom id when we encounter it. Then the increment starts counting from there.

src/Locations.py
```diff
@@ -17,7 +17,15 @@
...
 
-    location_table[key]["id"] = count
+    if "id" in location_table[key]:
+        item_id = location_table[key]["id"]
+        if item_id > count:
+            location_table[key]["id"] = item_id
+        else:
+            raise ValueError(f"{location_table[key]['name']} has an invalid ID. ID must be at least {count + 1}")
+    else:
+        location_table[key]["id"] = count
```
FuzzyGamesOn, line 27: Same note as above.

FuzzyGamesOn (changes requested):

src/Items.py
nicopop: > We might want to set count to the custom id when we encounter it. Then the increment starts counting from there.  

either that or we keep track of the custom ids somewhere and skip over them when counting up

silasary: Updating count is absolutely the intended behaviour, thanks for catching that

nicopop (approved):

nicopop (dismissed):
I dont think its as good as it could be since this wont allow the ID to go back at all but it should be good enough for now

src/Items.py
```diff
@@ -25,6 +25,13 @@
 
 # add sequential generated ids to the lists
 for key, val in enumerate(item_table):
+    if "id" in item_table[key]:
+        item_id = item_table[key]["id"]
+        if item_id > count:
```
axxroytovu, line 30: I don't think we should throw an error if `item_id == count` since that is just the default behavior.

axxroytovu: Also, is there any way to account for someone not putting the item IDs in order, but listing valid numbers? E.g. [1, 2, blank, 4, 3] will break, but could be solved by sorting the items to be [1, 2, 3, 4, blank].

axxroytovu (changes requested):

src/Items.py
FuzzyGamesOn: I'm not sure why someone would even want to have their custom ids out of order. That sounds like it should be something they should fix -- for their own sanity. 😜

silasary: If we allow people to go backwards, it means we need to do collision detection on every single increment, and that's much more expensive than just asking people to have things in order

axxroytovu (approved):

FuzzyGamesOn (approved):

PR #78, opened 2024-09-19
Remove table_processed hooks in favor of Data hooks

FuzzyGamesOn:
The `*_table_processed_hook` functions existed for Items.py, Locations.py, and Regions.py, and allowed power users to hook right before the items/locations/regions are processed individually after data import. However... there's nothing that happens between data import and the start of processing of those things. So the individual `*_table_processed_hook` functions are only different from the generic Data `after_load_*_file` functions in that they let you modify the tables in a non-persistent way (since init loads tables from Data) ... which leads to the table being out of sync with the lookup tables, which is probably always bad.

Also, the existence of these hooks often pushes hook devs to look at using these first instead of the Data hooks, which is not ideal (for above non-persistence reasons).

### tl;dr - It's more correct to use the Data.py hooks than to use these, so removing these.

If people are using these, the migration path is simply to move their hook code to the corresponding Data function instead.

### DIFF
--- a/src/Items.py
+++ b/src/Items.py
@@ -1,9 +1,7 @@
 from BaseClasses import Item
 from .Data import item_table
 from .Game import filler_item_name, starting_index
-from .hooks.Items import before_item_table_processed
 
-item_table = before_item_table_processed(item_table)
 
 ######################
 # Generate item lookups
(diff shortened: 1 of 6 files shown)


axxroytovu (approved):
Seems like a really good change. Cleans up a lot of duplicate hooks that were mostly just confusing.

silasary (approved):

PR #79, opened 2024-09-19
Client location buttons send on release instead of on press

FuzzyGamesOn:
This bug report inspired this change: https://discord.com/channels/1097532591650910289/1280257819584499892

From the bug report:
> As the title states, you can click on a location in the Manual client and it will send, but it can also send another location if another one winds up under your mouse cursor as you release the mouse button.

Currently, the buttons in the client use on_press, which is effectively a mousedown event. Instead, we want to use on_release, which is effectively a mouseup event. Tested with client and seems to work the same, just with probably fewer accidental clicks.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -288,7 +288,7 @@ class ManualContext(SuperContext):
                     self.death_link_button = Button(text="Death Link: Primed",
                                                 size_hint_x=None, width=150)
                     self.connect_layout.add_widget(self.death_link_button)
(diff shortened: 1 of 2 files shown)


nicopop (approved):
looks good

FuzzyGamesOn:
Realized I forgot to update client version. Oops 😄

nicopop (approved):

silasary (approved):

PR #80, opened 2024-09-19
Make JSON files identify their data type so a proper empty value is returned

FuzzyGamesOn:
So... I was doing weird stuff with a Manual where I deleted all the JSON files. And I got an error related to the pop() calls on the region and category table, because `load_data_file` returns `[]` whenever an exception is encountered.

So this PR makes each file indicate which type their contents should be, then they ensure that the default empty contents returned are of the type expected.

(Also, I could've moved `load_data_file` into the new class, but it seems useful as a utility method for people. And I'm also using it in my aforementioned Manual. 😃)

### DIFF
--- a/src/Data.py
+++ b/src/Data.py
@@ -20,6 +20,9 @@ def load_data_file(*args) -> dict:
     except:
         filedata = []
 
+        if args[0] in ['game.json', 'regions.json', 'categories.json', 'meta.json']:
+            filedata = {}
+
     return filedata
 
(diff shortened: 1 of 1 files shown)


nicopop (approved):
I know im not a requested reviewer but this works well with what im currently doing to try and implement verifying the data tables with their respective schema.
Thus why I tested this.

axxroytovu (approved):

PR #82, opened 2024-09-20
[DRAFT] Proof of concept: Validating data tables against their schema

nicopop:
# ill redo the description whenever im not tired

### DIFF
--- a/schemas/Manual.game.schema.json
+++ b/schemas/Manual.game.schema.json
@@ -81,8 +81,10 @@
                     }
                 },
                 "_comment": {"$ref": "#/definitions/comment"}
+
             },
-            "minProperties": 1
+            "minProperties": 1,
+            "additionalProperties": false
         }
     }
 }
--- a/src/Data.py
+++ b/src/Data.py
@@ -1,7 +1,5 @@
 import json
 import logging
-import os
-import pkgutil
 
 from .DataValidation import DataValidation, ValidationError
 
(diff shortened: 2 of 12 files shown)


FuzzyGamesOn:
I guess my main question is: Why would we want this? What problem is it solving, or purpose is it serving?

silasary:
Not a fan of this

silasary (changes requested):
To clarify my thoughts, I have two problems with this.

1. Hook developers can and do extend the json with extra data. The jsonschema does not like unknown fields, and will fail over the them.
2. This involves making multiple web requests during generation, often on the machine of a person who is neither the APWorld dev, nor the player playing the game. While this isn't technically a problem, some users will view it as a red flag, and might ban manuals from their games because of it. 

I do not want to merge this.

nicopop:
> To clarify my thoughts, I have two problems with this.
> 
> 1. Hook developers can and do extend the json with extra data. The jsonschema does not like unknown fields, and will fail over the them.
> 2. This involves making multiple web requests during generation, often on the machine of a person who is neither the APWorld dev, nor the player playing the game. While this isn't technically a problem, some users will view it as a red flag, and might ban manuals from their games because of it.
> 
> I do not want to merge this.
@silasary 
fair enough it was more of a proof of concept born out of an hyperfocus, there some stuff in there I liked and will probably recycle but the rest will go

silasary:
Yeah, it's really cool, and would make for a useful external tool. But it doesn't belong in DV

silasary:
Am I good to steal this code? I have a use for it :D

PR #84, opened 2024-09-22
Extend hint information

axxroytovu:
Add `extend_hint_information` and `before_extend_hint_information`

### DIFF
--- a/schemas/Manual.locations.schema.json
+++ b/schemas/Manual.locations.schema.json
@@ -92,6 +92,10 @@
                     "type": "boolean",
                     "default": false
                 },
+                "hint_entrance": {
+                    "description": "(Optional) Adds additional text to this location's hints to convey useful information. Typically used for entrance randomization.",
+                    "type": "string"
+                }
                 "_comment": {"$ref": "#/definitions/comment"}
             }
         },
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -31,7 +31,8 @@ from .hooks.World import \
     before_create_item, after_create_item, \
     before_set_rules, after_set_rules, \
     before_generate_basic, after_generate_basic, \
-    before_fill_slot_data, after_fill_slot_data, before_write_spoiler
+    before_fill_slot_data, after_fill_slot_data, before_write_spoiler, \
+    before_extend_hint_information
 from .hooks.Data import hook_interpret_slot_data
 
 class ManualWorld(World):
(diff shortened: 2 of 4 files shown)


axxroytovu:
Allows you to add "entrance" information to a hint if anyone wants to do that. 

See https://discord.com/channels/1097532591650910289/1097538232914296944/1287149476615557142

silasary:
Why not make it a json field?

axxroytovu:
It sounds like one of the main uses is in randomized situations, where the location would be read from the world state.

silasary:
Ah, yeah, that makes sense then

axxroytovu:
I ended up liking the json idea so I added that as an option

schemas/Manual.locations.schema.json
```diff
@@ -92,6 +92,10 @@
                     "type": "boolean",
                     "default": false
                 },
+                "hint_entrance": {
```
FuzzyGamesOn, line 95: Should probably have a name that more clearly says what it's for. Since it's a location name that's just for hinting, maybe something like `hint_name`.

src/__init__.py
```diff
@@ -332,6 +333,19 @@ def generate_output(self, output_directory: str):
     def write_spoiler(self, spoiler_handle):
         before_write_spoiler(self, self.multiworld, spoiler_handle)
 
+    def extend_hint_information(self, hint_data: dict[int, dict[int, str]]) -> None:
+        before_extend_hint_information(self, hint_data)
```
FuzzyGamesOn, line 337: Hooks should pass roughly the same set and order of arguments for consistency when possible, so either:
- self, self.multiworld, self.player (for hooks that aren't passing a collection to modify, like `create_regions`)
- [collection], self, self.multiworld, self.player (for hooks that are passing a collection to modify, like `create_items`)

So, hint data first, then the rest. 😃 Same for the hook below.

src/__init__.py
```diff
@@ -332,6 +333,19 @@ def generate_output(self, output_directory: str):
...
+        
+        for location in self.multiworld.get_locations(world.player):
```
FuzzyGamesOn, line 339: `world` should be `self`

FuzzyGamesOn (changes requested):

schemas/Manual.locations.schema.json
FuzzyGamesOn: > Considering it's shown under the "Entrance" column on the webtracker, I'm happy with the current name.

This also shows up in output from the `!hint` command in the client. 😃

axxroytovu: Even the `!hint` version maintains the `{location} in {world} at {entrance}` structure:

`[Hint]: CooltrainerTunic 's MP Berry x3 is at Shop - Coin 2 in AxxroyTunic 's World at Windmill Entrance . (not found)`

FuzzyGamesOn (approved):

FuzzyGamesOn:
PR only needs one more approval, so re-requesting reviews from everyone else. Last thing planned for upcoming unstable release.

PR #85, opened 2024-09-27
Filler placement at end of `create_items` should also use `get_filler_item_name()`

FuzzyGamesOn:
Realized from the upgrade steps that we're prepping for the unstable that `get_filler_item_name()` is only currently used for item links and `start_item_from_pool`. However, it doesn't really make sense to have the potential for the extras filler (that pads the item pool when they're short on items) to be different from the result of `get_filler_item_name()`.

So this PR makes the default addition of filler at the end of the create_items block in Manual also use `get_filler_item_name()` instead of using the game.json filler_item_name directly.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -105,7 +105,9 @@ class ManualWorld(World):
         configured_item_names = self.item_id_to_name.copy()
 
         for name in configured_item_names.values():
+            # victory gets placed via place_locked_item at the victory location in create_regions
             if name == "__Victory__": continue
+            # the game.json filler item name is added to the item lookup, so skip it until it's potentially needed later
             if name == filler_item_name: continue
 
             item = self.item_name_to_item[name]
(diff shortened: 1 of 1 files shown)


nicopop (approved):

src/__init__.py
```diff
@@ -355,8 +357,10 @@ def adjust_filler_items(self, item_pool, traps):
                 extra_item = self.create_item(self.random.choice(traps))
                 item_pool.append(extra_item)
 
+            filler_name_to_add = self.get_filler_item_name()
```
silasary, line 360: Can I get you to inline this? If people are using the hook, there's a reasonable chance it wants to return a different item each time it makes one, so we really want to request a name once per create_item.

silasary (changes requested):

src/__init__.py
axxroytovu: That was the intent of creating the hook

FuzzyGamesOn: There are three potential ways that people might want to update `filler_item_name`:

1. Updating `filler_item_name` before gen / needing nothing from world.
2. Updating `filler_item_name` to a single choice during gen and accounting for options, etc. from world.
3. Updating `filler_item_name` on each loop of Manual's filler padding during adjust_filler_items, also accounting for options, etc. from world.

1 is separate from `get_filler_item_name()` (and has been supported for a long time anyways). 2 and 3 are both potential use cases for `get_filler_item_name()` in relation to Manual's padding of filler items. So I updated the PR to support 2 and 3 both by making `filler_item_name` a prop on world that people can update via other hooks (for 2).

nicopop (approved):

silasary (approved):
Yup, I'm happy with this now

PR #86, opened 2024-09-28
Game Select dropdown

silasary:
image
image

Also fixes a bug when launching the client from the Command Line.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -280,8 +280,8 @@ class ManualContext(SuperContext):
                 manuals = [w for w in AutoWorldRegister.world_types.keys() if "Manual_" in w]
                 manuals.sort()  # Sort by alphabetical order, not load order
                 self.manual_game_layout.add_widget(game_bar_label)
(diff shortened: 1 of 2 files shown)


nicopop:
How does it look and work if you have way too many manual installed?
Leaving the text input to work as a search field might be a good idea.

FuzzyGamesOn:
We should probably tweak this a bit visually, too. The spinner options when shown have an inconsistent height compared to the spinner, and they're also the same color as the spinner (and the tabs underneath). We could probably go with a bit darker background on the options to visually distinguish them.

nicopop:
small issue with this if tracker is on thats what is loaded instead of manual client

silasary:
I agree with both of these. Kivy does not make either easy.

silasary:
Made them shorter, sorted alphabetically, and slightly darker.
image

nicopop (approved):

nicopop (approved):
looks good, ill look into kivy search for my idea but as is this works

FuzzyGamesOn (approved):

PR #89, opened 2024-09-28
Add feature to automatically send checks when UT says they're in logic

silasary:
Alternative take on #74.
Not tested yet, do not merge.

### DIFF
--- a/schemas/Manual.game.schema.json
+++ b/schemas/Manual.game.schema.json
@@ -42,6 +42,10 @@
             "type": "integer",
             "default": 1
         },
+        "auto_send": {
+            "description": "(Optional) Will your client automatically send in-logic checks? WARNING: This should only be used for non-game manual worlds.",
+            "type": "boolean"
+        },
         "_comment": {"$ref": "#/definitions/comment"}
     },
     "required":["game", "filler_item_name"],
--- a/src/Game.py
+++ b/src/Game.py
@@ -14,3 +14,5 @@ if "starting_index" in game_table:
         raise Exception("The value of data/game.json:'starting_index' should be an int")
 else:
     starting_index = 1
+
+auto_send = game_table["auto_send"] if "auto_send" in game_table else False
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -50,7 +50,11 @@ class ManualClientCommandProcessor(ClientCommandProcessor):
             self.output(response)
             return False
 
-
+    def _cmd_toggle_autosend(self) -> bool:
+        """Toggle autosend.  If enabled, the client will automatically send checks out as soon as UT says they're in logic."""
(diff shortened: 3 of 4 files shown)


FuzzyGamesOn:
I don't think this should be a Game option. Features that auto-spam a multiworld without requiring user interaction first are probably a bad idea in general. 

imo, this should always be an opt-in that's toggled on after connecting in the client. A button to toggle would be fine, or a command, either works.

silasary:
I made it a game option because LinkLink and Progressivizer pretty much always want it enabled, and that's what the original PR did

nicopop:
if we want to "hide" it we could always place it in the meta.json since the stuff in that file is more technical

axxroytovu (approved):
I'm in favor of the current setup. It feels like allowing the game author to set the auto-send to "default on" which feels perfectly reasonable for the more meta uses of manual.

FuzzyGamesOn:
We're not just designing a feature for a handful of meta apworlds to use. We're designing a feature that *any* Manual apworld can use. And this current design of the feature is one that could be enabled **silently without the player of that world knowing**. 

**That's really bad.** And I'm shocked that I have to (again) mention that that's bad.

Also, this current design of the feature has the potential for large-scale location spam, which is a concern I brought up in the original feature suggestion thread for the feature: https://discord.com/channels/1097532591650910289/1181665955542212740/1181699938086817842

We have to consider the main potential risks of this feature when we're making it. Please revise the feature in a way that makes it safe for users and for multiworlds, or close the PR.

axxroytovu:
Clarifying question: what do you mean by "enabled silently without the player of that world knowing"?

Are you saying that people will enable autorelease on other peoples' worlds without their knowledge, or that the author enabling a feature is "without the player's knowledge"?

axxroytovu:
Is there a way to add a "two-factor" check similar to deathlink? I.e. t must be turned on both in the game options and the player yaml to be enabled by default. Or is that just not worth the extra hassle and it's better to have the player type `/toggle_autosend` every time?

FuzzyGamesOn:
> Clarifying question: what do you mean by "enabled silently without the player of that world knowing"?
> 
> Are you saying that people will enable autorelease on other peoples' worlds without their knowledge, or that the author enabling a feature is "without the player's knowledge"?

The latter: that the author enabling a feature is "without the player's knowledge".

If I'm a player, and a world creator releases an apworld for a game franchise that I know/enjoy, and I decide to take their apworld into a sync/async... What indication do I have (as a player) that the world creator enabled this option in their JSON configs? And that, when I connect for the first time, whatever they marked as open is going to get automatically sent into the multiworld I joined.

silasary:
I'm just going to point people at https://github.com/gjgfuj/AP-SlowRelease/releases if they want this functionality

PR #90, opened 2024-09-29
For some reason, we were still starting Location IDs at 501.  Let's not do that.

silasary:


nicopop (approved):
Oops I missed that lol

silasary:
Yeah, I only noticed it when implementing ID skips in XIV, and it was like "50 is lower than 526"

axxroytovu (approved):

FuzzyGamesOn:
We'll probably want to do another unstable release with this in it, so people can test the new ids without this bug. Good find, btw! 😃

FuzzyGamesOn (approved):
Yeah, this probably should've been changed a long time ago. Not sure it was ever needed 😃

silasary:
It wasn't :D

PR #91, opened 2024-09-29
Rule functions recursion

nicopop:
I got inspired by the problem someone had on discord because of the fact we only ran functions once and at the time said it would be complicated to fix it... 
So here's a PR adding recursion to the requires's function finding and executing.
I included a maximum recursion depth param that can be changed (currently set to 5)

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -78,27 +78,37 @@ def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
         if requires_list == "":
             return True
 
-        for item in re.findall(r'\{(\w+)\((.*?)\)\}', requires_list):
-            func_name = item[0]
-            func_args = item[1].split(",")
-            if func_args == ['']:
-                func_args.pop()
+        def findAndRecursivelyExecuteFunctions(requires_list: str, maxRecursion: int, recursionDepth: int = 0) -> str:
+            found_functions = re.findall(r'\{(\w+)\((.*?)\)\}', requires_list)
+            if found_functions:
+                if recursionDepth >= maxRecursion:
+                    raise RecursionError(f'the require in {area.get("name", f"An area with these parameters: {area}")} looped too many time ({maxRecursion})')
+                else:
+                    for item in found_functions:
+                        func_name = item[0]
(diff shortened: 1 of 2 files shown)


nicopop:
was tested using these:
```python
# You can also return a string from your function, and it will be evaluated as a requires string.
def recursionTest(world: World, multiworld: MultiWorld, state: CollectionState, player: int):
    """Returns a requires string that checks if the player has unlocked the tank."""
    return "{hello()}"
def recursionFail(world: World, multiworld: MultiWorld, state: CollectionState, player: int):
    """Returns a requires string that checks if the player has unlocked the tank."""
    return "{recursionFail()}"
def hello(world: World, multiworld: MultiWorld, state: CollectionState, player: int):
    return True
```

src/Rules.py
```diff
@@ -78,27 +78,39 @@ def checkRequireStringForArea(state: CollectionState, area: dict):
...
+                        result = func(world, multiworld, state, player, *func_args)
+                        if isinstance(result, bool):
+                            requires_list = requires_list.replace("{" + func_name + "(" + item[1] + ")}", "1" if result else "0")
+                        else:
+                            requires_list = requires_list.replace("{" + func_name + "(" + item[1] + ")}", str(result))
+
+                requires_list = findAndRecursivelyExecuteFunctions(requires_list, maxRecursion, recursionDepth + 1)
+            return requires_list
 
+        requires_list = findAndRecursivelyExecuteFunctions(requires_list, maxRecursion=5)
```
FuzzyGamesOn, line 113: Probably better to hardcode max recursion instead of passing it in, either in the function or on the world. That way, anyone reusing this later can't accidentally screw up their max recursion and introduce an infinite loop.

I doubt we want a customizable max per run, anyways.

FuzzyGamesOn (changes requested):
Minor change

src/Rules.py
nicopop: ill put the variable on the world so dev can change it via hooks if they really wants even if more recursion = more generation time

FuzzyGamesOn (approved):

silasary (approved):

PR #95, opened 2024-10-15
Add CSV support

silasary:
This is terrifying, and not the greatest.  However, it is very good for simple things, and I think a lot of people would get value out of it, especially when entering repetitive data.

This PR adds:
* Support for `items.csv` and `locations.csv` (The former is much more useful, but I'm not gonna stop people from using the latter)
* `load_data_csv` helper function.  Hooks users will probably find this useful
* Finally adds a bugfix for when your category is a string rather than a list (Because CSV items will always be a single string)
* Option to use `"classification": "progression"` in your items.  This lets you have a single column, rather than five.

This could definitely be improved.  Off the top of my head:
* Special logic to lowercase headings/values that should be lowercase.
* Support for multiple categories
* Data Validation?

I wrote this on the train ride home from PAX, and was very sleep deprived at the time.  Do we even want it?

### DIFF
--- a/src/Data.py
+++ b/src/Data.py
@@ -1,7 +1,8 @@
 import logging
+from typing import Literal
 
 from .DataValidation import DataValidation, ValidationError
-from .Helpers import load_data_file as helpers_load_data_file
+from .Helpers import load_data_file as helpers_load_data_file, load_data_csv
 
 from .hooks.Data import \
     after_load_game_file, \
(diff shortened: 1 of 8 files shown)


FuzzyGamesOn:
At some point, yeah, we want support for alternate formats. Our documentation needs to catch up before we do that, though, so people have a point of reference for the options regardless of format.

The classification option does seem useful on its own, though. Might be worth adding that in a separate PR in the meantime.

Also, once we get to adding alternate formats, Builder would need to be updated to support them.

silasary:
Closing in favor of #96 #97 #98

PR #96, opened 2024-10-16
If provided with a string for category, don't iterate over each letter

silasary:


FuzzyGamesOn:
This seems good. We're technically allowing unsupported syntax when we do this, though. So we have to do one of two things:

1. Warn the user that their syntax is incorrect while we're making it work anyways.
2. Update the schema to indicate that category can now be either a string or an array.

FuzzyGamesOn:
(Also, please add a couple reviewers to your PRs so we have active reviewers on them. And since one of the reviewers has to be an admin, add me to each. :smiley: )

FuzzyGamesOn (approved):

nicopop (approved):
looks good.

PR #97, opened 2024-10-16
Classification field, as an alternative to boolean fields for each

silasary:


FuzzyGamesOn:
Easy enough. We'll also need a related PR in Builder to support users importing this.

nicopop:
Don't forget to add that to the json schema

nicopop (approved):

FuzzyGamesOn:
I can't approve this until there's a Builder PR to support the classification field, since using it instead of the existing classification keys would cause worlds to not import correctly to Builder.

silasary:
Weirdly enough, the builder already supports the classification field.

Internally, the builder uses item["classification"] instead of all the bools.  And the way it loads the items, if that field is prepopulated, it'll just roll with the value.

A round trip through the builder converts the item to the old format, but it works well enough for my liking.

FuzzyGamesOn:
It's awesome that it sorta supports it, though I somewhat disagree on the quality bar. But that's kinda irrelevant because I realized...

#105 added support for multiiple item classifications, which I believe was started and completed after this PR was started. So this PR would need to be updated to similarly support multiple item classifications to ensure that it can have the same information as the separate fields.

Also realized we didn't get #105 to update Builder to support it. So no reason I should require any level of support for this one, either. Just feature parity in core is fine.

---

I swear I'm not trying to find things to hold this PR up. Just unfortunate timing on this one :sweat_smile:

FuzzyGamesOn:
Ran back through this because of the label. Looks like it's good to run with once it supports multiple values, to be consistent with the current multiple classification support in separate keywords.

src/__init__.py
```diff
@@ -282,6 +283,9 @@ def create_item(self, name: str, class_override: Optional['ItemClassification']=
         else:
             classification = ItemClassification.filler
 
+        if "classification" in item and item["classification"]:
+            classification = reduce((lambda a, b: a | b), {ItemClassification[str_classification.strip()] for str_classification in item["classification"].split(",")})
+
```
nicopop: missing and else: here so the other base classification is never run currently
```suggestion
        if "classification" in item and item["classification"]:
            classification = reduce((lambda a, b: a | b), {ItemClassification[str_classification.strip()] for str_classification in item["classification"].split(",")})
        else:
```

nicopop (changes requested):
### You forgot this important part:
You need to make some change in datavalidation because [checkItemsThatShouldBeRequired()](https://github.com/ManualForArchipelago/Manual/blob/8b80a40f6dfa3fe4d93375550a672e2277431195/src/DataValidation.py#L211)
will not see item created with this new property as valid progs, and thus unit tests will fail and thus gen fails
there's a reason why I added classification_count to it
To prove this all I had to do is make 
```json

        {
            "name": "Shuma-Gorath",
            "category": [
                "Characters",
                "Right Side"
            ],
            "value": {"star": 5},
            "classification": "progression",
            "sort-key": "row-0-8"
        },
```
and now gen fail to... gen
edit:
just add the following on line 221 of Datavalidation.py:
\+ add import for reduce of course
```python

            # if the string classification contains progresion then no check needed
            if item.get("classification"):
                classification = reduce((lambda a, b: a | b), {ItemClassification[str_classification.strip()] for str_classification in item["classification"].split(",")})
                if ItemClassification.progression in classification:
                    continue

```

src/__init__.py
silasary: Ah, yep.  That was added in a merge resolution and I didn't notice the logical implications.  Easy to fix 😁

nicopop:
> ### You forgot this important part:
> You need to make some change in datavalidation because [checkItemsThatShouldBeRequired()](https://github.com/ManualForArchipelago/Manual/blob/8b80a40f6dfa3fe4d93375550a672e2277431195/src/DataValidation.py#L211) will not see item created with this new property as valid progs, and thus unit tests will fail and thus gen fails there's a reason why I added classification_count to it
> To prove this all I had to do is make
> 
> ```json
>         {
>             "name": "Shuma-Gorath",
>             "category": [
>                 "Characters",
>                 "Right Side"
>             ],
>             "value": {"star": 5},
>             "classification": "progression",
>             "sort-key": "row-0-8"
>         },
> ```
> 
> You get this:
> image
> edit: just add the following on line 221 of Datavalidation.py: + add import for reduce of course
> 
> ```python
>             # if the string classification contains progresion then no check needed
>             if item.get("classification"):
>                 classification = reduce((lambda a, b: a | b), {ItemClassification[str_classification.strip()] for str_classification in item["classification"].split(",")})
>                 if ItemClassification.progression in classification:
>                     continue
> ```

### Pretty sure this is still an issue, you would see it if you add an example usage of this PR in items.json (which you probably should)

nicopop (approved):

PR #98, opened 2024-10-16
Helper function for loading a csv file.  People might find this useful for hooks

silasary:


src/Helpers.py
```diff
@@ -1,8 +1,9 @@
+import csv
 import os
 import pkgutil
 import json
 
-from BaseClasses import MultiWorld, Item, Location
```
FuzzyGamesOn, line 5: ManualLocation comes from Location and is still used below, so we need to add this back.

FuzzyGamesOn (changes requested):
Just the one fix, everything else looks good.

src/Helpers.py
silasary: It's not used at all in this file

FuzzyGamesOn: Oh, you're right. Misread, carry on :smiley:

nicopop (approved):
Looks good.

FuzzyGamesOn (approved):

PR #99, opened 2024-11-02
Backport 0.5.1's run_gui function 

silasary:
This was the easiest way to do it

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -213,6 +213,18 @@ class ManualContext(SuperContext):
         if events:
             self.ui.update_tracker_and_locations_table(update_highlights=True)
 
+    def run_gui(self):
+        """Import kivy UI system from make_gui() and start running it as self.ui_task."""
+        if hasattr(SuperContext, "make_gui"):
+            # Call the real one if it exists
+            return super().run_gui()
+
+        # This is a copy of 0.5.1's run_gui, because backporting is easier than the alternative.
+        # This entire function can be removed once 0.5.1 is the old enough.
+        ui_class = self.make_gui()
+        self.ui = ui_class(self)
+        self.ui_task = asyncio.create_task(self.ui.async_run(), name="UI")
+
     def make_gui(self) -> typing.Type["kvui.GameManager"]:
         ui = super().make_gui()  # before the kivy imports so kvui gets loaded first
 


nicopop (approved):
with my limited knowledge of clientside I can say this looks valid

axxroytovu (approved):

FuzzyGamesOn (approved):

src/ManualClient.py
```diff
@@ -213,6 +213,18 @@ def on_tracker_events(self, events: list[str]):
         if events:
             self.ui.update_tracker_and_locations_table(update_highlights=True)
 
+    def run_gui(self):
+        """Import kivy UI system from make_gui() and start running it as self.ui_task."""
+        if hasattr(SuperContext, "make_gui"):
```
axxroytovu, line 218: How did we all miss that this if statement is backwards. We're checking if the context has `make_gui` then running the opposite

FuzzyGamesOn: I read this as: If the SuperContext has make_gui (i.e., we detected that we're in 0.5.1 RC or beyond), just call the parent's run_gui instead of doing our custom run_gui stuff. Since our custom run_gui is only needed for prior to RC. RC will call our make_gui instead of run_gui and expect a UI in response, so we don't want to override the parent run_gui anymore in that case.

Then, since we initially moved our old run_gui code into make_gui, anything prior to RC will -- via this backported run_gui method -- just call our stuff in make_gui (which was previously our run_gui code) to get a UI object, then run the old run_gui async.

I haven't tested the code, though, so can't attest whether it fixes the issue or not. Just appears to.

silasary: Fuzzy Is correct. 
If CommonContext has a make_gui, it's 0.5.1+, and we run it's run_gui. 
If it doesn't have it, we run the below code, which is a copy of 0.5.1RC1's run_gui.

The reason I did it this way is because it doesn't override potential bugfixes or upgrades to the function in the future.

axxroytovu: Attempting to actually make this on 0.5.0 gives this error
```
Traceback (most recent call last):
  File "multiprocessing\process.py", line 314, in _bootstrap
  File "multiprocessing\process.py", line 108, in run
  File "C:\ProgramData\Archipelago\custom_worlds\src.apworld\src\ManualClient.py", line 758, in launch
    asyncio.run(main(args))
  File "asyncio\runners.py", line 190, in run
  File "asyncio\runners.py", line 118, in run
  File "asyncio\base_events.py", line 653, in run_until_complete
  File "C:\ProgramData\Archipelago\custom_worlds\src.apworld\src\ManualClient.py", line 734, in main
    ctx.run_gui()
  File "C:\ProgramData\Archipelago\custom_worlds\src.apworld\src\ManualClient.py", line 224, in run_gui
    ui_class = self.make_gui()
               ^^^^^^^^^^^^^^^
  File "C:\ProgramData\Archipelago\custom_worlds\src.apworld\src\ManualClient.py", line 229, in make_gui
    ui = super().make_gui()  # before the kivy imports so kvui gets loaded first
         ^^^^^^^^^^^^^^^^
AttributeError: 'super' object has no attribute 'make_gui'
```

axxroytovu: Manual build was on source as of today
Universal Tracker is v0.1.11 - the latest build that is functional on 0.5.0

silasary: Ah. yep.  That's on me for not testing well enough.

PR #101, opened 2024-11-18
WIP Itemvalue misc fixes

nicopop:
includes multiple small fixes to ItemValue functionality:
2586167 Fixed datavalidation to only check for requirements of used regions 
63605b3 Made it so for sure that value name get lowercased (check items.py)
2b97f9d Renamed world.item_values_cache to world.itemvalue_rule_cache to stop potential confusion with world.item_values 
5e64593 Stopped PreFill Datavalidation of ItemValue from reseting the world.item_values cache when running get_items_with_value so hook user can run it first then edit it to modify an item value for dev reasons 

for some reason the diff of  commit 63605b3 failed the only line changed in datavalidation is [249 -> 278](https://github.com/ManualForArchipelago/Manual/compare/ItemValue-Misc-fixes?expand=1#diff-9e2dd0bc6d4ea837c08d3aabd6126fa53d7ff1982b72ceef74b4f548502144cfR278)

PR #103, opened 2024-11-18
Make Item values all be lowercase

nicopop:
as sugested in #101 it was split in its own PR Still need to find why removing the .lower.strip from key in datavalidation make the diff think I changed the entire file...

to help with some bug in datavalidation make it so values are actually lowercase in the dict
before this most of the ItemValue code did name.lower.strip() everytime

### DIFF
--- a/src/Items.py
+++ b/src/Items.py
@@ -47,8 +47,12 @@ for item in item_table:
             item_name_groups[c] = []
         item_name_groups[c].append(item_name)
 
+    #Just lowercase the values here to remove all the .lower.strip down the line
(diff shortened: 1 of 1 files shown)


nicopop:
An alternative to this pr is to remove all the .lower() in everything related to ItemValue (which is a possibility if you all feel like it's a better idea)
When I originally made ItemValue, I didn't want devs to make value type whose only difference was capitalisation (star vs Star or Coin vs coin)

silasary (approved):
I'm happy with this

FuzzyGamesOn (approved):

PR #104, opened 2024-11-18
Itemvalue misc fixes

nicopop:
_I told myself I should stop making PR past midnight but here's one._
# this is #101 but the diff work this time
og. lowercase commit was split into #103 
includes multiple small fixes to ItemValue functionality:
b21f576 Renamed world.item_values_cache to world.itemvalue_rule_cache to stop potential confusion with world.item_values
88854ef Split/changed the force reset parameter of the cache of get_items_with_value to be a skip instead. Mainly so it work more like the ItemValue rule function. I made some helper.py functions to replace the reset usage
5150bbc Made both datavalidation of ItemValue (prefill and assert) only check for region with locations in it and their parents using some recursion tech

## to explain the how/why of 5150bbc
my fix for https://discord.com/channels/1097532591650910289/1307802587839594617
### WHY: 
**_Hooks and Options can remove locations._**  
I could easily add a check to only test for region with locations but its still possible for a region without to be logically required since parent region can be empty.
### HOW:
it work basicaly the same for the PreFill test and the Assert one
the only difference is the way to find the parents
- ~~**Assert:** Since we only have the basic json, I build a dict(child,list(Parent)) that way I can do the recursion below~~ Removed
- **Prefill:** I could probably also use the json but I decided to use the entrances since multiworld did it for me anyway  
#### in both case:
1.  it checks every region for locations, and if present, add the region to a list
2.  since a region can be empty but still be a parent of any of the region on the "has  Loc" list, Check recursively the parents and add them to the list.
    2.1 add the parents to an "already checked" list so it's skipped in future recursion
3. once all the used region are identified check their requires, entrance_require and exit require for ItemValue (if the region on the otherside is used of cource).
4. check every location of the region for ItemValues requirement
5. Finally check if we have the items to fullfill those requirements

### DIFF
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -223,15 +223,22 @@ class DataValidation():
 
         used_regions = filter_used_regions_for_player(world, player_regions, player)
 
+        used_regions_names = {r.name for r in set(used_regions)}
         used_regions = {r for r in set(used_regions) if r.name not in ['Manual', 'Menu']}#Manual shouldn't have any requires so skip it
         #Check used regions (and their parent(s)) for ItemValue requirement
         for region in used_regions:
             manualregion = DataValidation.region_table.get(region.name, {})
             if manualregion:
                 if "requires" in manualregion and manualregion["requires"]:
-                    region_requires = json.dumps(manualregion["requires"])
+                    DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, json.dumps(manualregion["requires"]))
 
-                    DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, region_requires)
+                for region_entrance, require in manualregion.get('entrance_requires', {}).items():
+                    if region_entrance in used_regions_names:
+                        DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, json.dumps(require))
+
+                for region_exit, require in manualregion.get('exit_requires', {}).items():
+                    if region_exit in used_regions_names:
+                        DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, json.dumps(require))
 
 
                 for location in region.locations:
--- a/src/DataValidation.py
+++ b/src/DataValidation.py
@@ -211,6 +211,7 @@ class DataValidation():
     def checkIfEnoughItemsForValue():
         values_available = {}
         values_requested = {}
+        used_regions = set(["Manual"])
 
         # First find the biggest values required by locations
         for location in DataValidation.location_table:
@@ -219,10 +220,38 @@ class DataValidation():
 
             # convert to json so we don't have to guess the data type
             location_requires = json.dumps(location["requires"])
-
+            if location.get('region'):
+                used_regions.add(location['region'])
             DataValidation._checkLocationRequiresForItemValueWithRegex(values_requested, location_requires)
-        # Second, check region requires for the presence of item name
+
+        parent_child = {}
+        # First build a parent-child dictionary:
         for region_name in DataValidation.region_table:
+            region = DataValidation.region_table[region_name]
+
+            region_connect_to = region.get("connects_to", [])
+            for child in region_connect_to:
+                if child not in parent_child.keys():
+                    parent_child[child] = []
+                parent_child[child].append(region_name)
+
+        # add parent and their parent to used_regions recursively if any location is present
+        checked_parent = ['Manual']
+        for region_name in set(used_regions):
+            def checkParent(name):
+                if name in checked_parent: #dont check a region twice
+                    return
+                checked_parent.append(name)
+                if name in parent_child.keys():
+                    for parent in parent_child[name]:
+                        checkParent(parent)
+                        used_regions.add(parent)
+                return
+            checkParent(region_name)
+
+        used_regions.remove('Manual')
+        # Second, check region requires of used regions for the presence of ItemValue
+        for region_name in used_regions:
             region = DataValidation.region_table[region_name]
 
             if "requires" not in region:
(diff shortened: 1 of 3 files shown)


nicopop:
I was waiting for some ItemValue bug report before undrafting this PR but it seems that report wont come so ill mark this as ready

src/DataValidation.py
FuzzyGamesOn, line 1: The changes in this file depend on answering a question I brought up in my previous PR that turned off this pre-fill check. From that PR:
> So this PR comments out the pre_fill validation for ItemValue -- to fix the bugs with category disabling -- until we figure out what we want to do with it. (And when we're figuring out what to do with it, we should figure out if -- and why -- we want pre_fill validation.)

I did see the "why" in the PR description, but that seemed to be justifying one solution over another. I'm mainly asking why we want a second validation step at pre_fill for ItemValue, when we have the stage_assert one already.

I feel like the answer might be "to protect users from their own hook changes", and I'm not in favor of adding validation for that. Because of how open-ended hooks are, that's a losing battle. But I'm open to other reasons/ideas. :smiley:

nicopop: its not just for hooks, stage assert only see and test using the original data jsons, any changes (missing items and/or locations) from our own option system will be missed there. without the test in prefill it leads to some generic generation failure error.
Now that I think about it we definetly only need 1 of those, I personally would prefer we keep the Prefill one for the reason written above.

FuzzyGamesOn: Ah, yeah, if something built-in can cause it, that's a great reason to have it. :smiley: Also agree on keeping the pre_fill one over the assert_stage one in that case. Also cuts down on the used_region check duplication.

FuzzyGamesOn (changes requested):

src/DataValidation.py
nicopop: ill remove the stage_assert one in a commit soon

src/Helpers.py
```diff
@@ -128,10 +128,21 @@ def get_items_for_player(multiworld: MultiWorld, player: int, includePrecollecte
...
-def get_items_with_value(world: World, multiworld: MultiWorld, value: str, player: Optional[int] = None, force: bool = False) -> dict[str, int]:
+def reset_specific_item_value_cache_for_player(world: World, value: str, player: Optional[int] = None) -> dict[str, int]:
+    if player is None:
+        player = world.player
+    return world.item_values[player].pop(value, {})
+
+def reset_item_value_cache_for_player(world: World, player: Optional[int] = None):
+    if player is None:
+        player = world.player
+    world.item_values[player] = {}
```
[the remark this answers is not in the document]
nicopop: the goal was to give hook devs a replacement way to reset the cache since I removed the forced option but I forgot to remove the docstring comment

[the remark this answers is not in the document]
nicopop: would you prefer that I keep the functions (and adjusted docstring) or just have the docstring?

src/DataValidation.py
```diff
@@ -263,12 +292,37 @@ def preFillCheckIfEnoughItemsForValue(world: World, multiworld: MultiWorld):
...
+                region = player_regions[name]
+                if name in checked_parent: #dont check a region twice
+                    return
+                checked_parent.append(name)
+                if region.entrances:
+                    for entrance in region.entrances:
+                        checkParent(entrance.parent_region.name)
+                        used_regions.add(entrance.parent_region.name)
+                return
+            checkParent(region_name)
```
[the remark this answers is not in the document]
nicopop: function moved [here](https://github.com/ManualForArchipelago/Manual/pull/104/files#diff-29c9513a0e849f4eb01a0a7bf4bef077df0f184ee4cc3d300daf8e41d8348502R173-R208)
I tried to let the function get the regions by itself but it caused a memory leak so instead its just a filtering

src/Helpers.py
```diff
@@ -143,15 +154,49 @@ def get_items_with_value(world: World, multiworld: MultiWorld, value: str, playe
...
+    if value not in world.item_values.get(player, {}).keys() or skipCache:
         item_with_values = {i.name: world.item_name_to_item[i.name]['value'].get(value, 0)
                             for i in player_items if i.code is not None
                             and i.name in world.item_name_groups.get(f'has_{value}_value', [])}
+        if skipCache:
+            return item_with_values
         world.item_values[player][value] = item_with_values
     return world.item_values[player].get(value)
+
+def filter_used_regions_for_player(world: World, player_regions: dict|list, player: Optional[int] = None, ) -> set:
```
FuzzyGamesOn, line 173: Trailing comma

FuzzyGamesOn (changes requested):
Minor typo, everything else seems good

src/Helpers.py
nicopop: That's from when I moved player_regions from optional to instead be required.(looping the multiworld regions in that function somehow created a potential memory leak warning from ap)
I'll fix that when I'm back to my PC.

nicopop: neither player or world are actually used now that the multiworld.regions stuff is done outside of it so I removed them at the same time I fixed this typo

FuzzyGamesOn (approved):

src/DataValidation.py
```diff
@@ -1,519 +1,480 @@
-import logging
```
silasary, line 1: What's going on with the diff here?

silasary: Oh, we went from Windows line endings to Linux ones (`\r\n` to `\n`)

silasary (dismissed):

src/DataValidation.py
nicopop: Idk why, since im on Windows...

FuzzyGamesOn (dismissed):
Can't believe nico dismissed my previous review, smh my head

silasary (approved):

FuzzyGamesOn (approved):

PR #105, opened 2024-11-18
Add support for assigning multiple item classifications

Seatori:
Adds support for users assigning multiple item classifications to a single item.

This is achieved by taking the base `filler` flag and appending each of the assigned classifications to it, making use of how Manual already 'supported' assigning multiple classifications to an item, but this adjusts the behavior so that they're no longer mutually exclusive.

Tested by assigning all three of the "Progression", "Useful", and "Trap" flags to a single item and then viewing that it displayed with all three in the client. Also verified that Useful + Trap items were correctly removed from the pool when the number of items exceeded the number of locations.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -208,24 +208,18 @@ class ManualWorld(World):
         name = before_create_item(name, self, self.multiworld, self.player)
 
         item = self.item_name_to_item[name]
-        filler = ItemClassification.filler
-        trap = ItemClassification.filler
-        useful = ItemClassification.filler
-        progression = ItemClassification.filler
-        progression_skip_balancing = ItemClassification.filler
+        classification = ItemClassification.filler
 
         if "trap" in item and item["trap"]:
-            trap = ItemClassification.trap
+            classification |= ItemClassification.trap
 
         if "useful" in item and item["useful"]:
-            useful = ItemClassification.useful
+            classification |= ItemClassification.useful
 
-        if "progression" in item and item["progression"]:
-            progression = ItemClassification.progression
-        elif "progression_skip_balancing" in item and item["progression_skip_balancing"]:
-            progression_skip_balancing = ItemClassification.progression_skip_balancing
-
-        classification = filler | trap | useful | progression | progression_skip_balancing
+        if "progression_skip_balancing" in item and item["progression_skip_balancing"]:
+            classification |= ItemClassification.progression_skip_balancing
+        elif "progression" in item and item["progression"]:
+            classification |= ItemClassification.progression
 
         item_object = ManualItem(name, classification,
                         self.item_name_to_id[name], player=self.player)
@@ -392,9 +386,15 @@ class ManualWorld(World):
             fillers = [item for item in item_pool if item.classification == ItemClassification.filler]
             traps = [item for item in item_pool if item.classification == ItemClassification.trap]
             useful = [item for item in item_pool if item.classification == ItemClassification.useful]
+            # Useful + Trap is classified separately so that it can have a unique priority ranking.
+            useful_traps = [item for item in item_pool if
+                            ItemClassification.progression not in item.classification
+                            and ItemClassification.useful in item.classification
+                            and ItemClassification.trap in item.classification]
             self.random.shuffle(fillers)
             self.random.shuffle(traps)
             self.random.shuffle(useful)
+            self.random.shuffle(useful_traps)
             for _ in range(0, abs(extras)):
                 popped = None
                 if fillers:
@@ -403,6 +403,9 @@ class ManualWorld(World):
                     popped = traps.pop()
                 elif useful:
                     popped = useful.pop()
+                # Not sure if Useful + Trap should go before or after Useful.
+                elif useful_traps:
+                    popped = useful_traps.pop()
                 else:
                     logging.warning("Could not remove enough non-progression items from the pool.")
                     break
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -208,19 +208,24 @@ class ManualWorld(World):
         name = before_create_item(name, self, self.multiworld, self.player)
 
         item = self.item_name_to_item[name]
-        classification = ItemClassification.filler
+        filler = ItemClassification.filler
+        trap = ItemClassification.filler
+        useful = ItemClassification.filler
+        progression = ItemClassification.filler
+        progression_skip_balancing = ItemClassification.filler
 
         if "trap" in item and item["trap"]:
-            classification = ItemClassification.trap
+            trap = ItemClassification.trap
 
         if "useful" in item and item["useful"]:
-            classification = ItemClassification.useful
+            useful = ItemClassification.useful
 
         if "progression" in item and item["progression"]:
-            classification = ItemClassification.progression
+            progression = ItemClassification.progression
+        elif "progression_skip_balancing" in item and item["progression_skip_balancing"]:
+            progression_skip_balancing = ItemClassification.progression_skip_balancing
 
-        if "progression_skip_balancing" in item and item["progression_skip_balancing"]:
-            classification = ItemClassification.progression_skip_balancing
+        classification = filler | trap | useful | progression | progression_skip_balancing
 
         item_object = ManualItem(name, classification,
                         self.item_name_to_id[name], player=self.player)
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -383,6 +383,9 @@ class ManualWorld(World):
                 item_pool.append(extra_item)
         elif extras < 0:
             logging.warning(f"{self.game} has more items than locations. {abs(extras)} non-progression items will be removed at random.")
+            # Filler is only assigned if the item doesn't have any other tags, so it only has to be covered by itself.
+            # Skip Balancing is also not covered due to how it's only supported when paired with Progression.
+            # As a result, these cover every possible combination can be removed.
             fillers = [item for item in item_pool if item.classification == ItemClassification.filler]
             traps = [item for item in item_pool if item.classification == ItemClassification.trap]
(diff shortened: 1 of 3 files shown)


Seatori:
As of recent, assigning multiple classifications to items has been added to core a few times, and it's considered to be supported behavior. Manual technically already allows multiple classifications to be assigned to an item, but it overwrites them in sequence. This attempts to adjust the behavior so that it treats them as additive instead.

src/__init__.py
```diff
@@ -208,19 +208,24 @@ def create_item(self, name: str) -> Item:
         name = before_create_item(name, self, self.multiworld, self.player)
 
         item = self.item_name_to_item[name]
-        classification = ItemClassification.filler
+        filler = ItemClassification.filler
+        trap = ItemClassification.filler
+        useful = ItemClassification.filler
+        progression = ItemClassification.filler
+        progression_skip_balancing = ItemClassification.filler
```
FuzzyGamesOn, line 215: If you're applying filler to start, makes more sense to just apply flags as you go instead of accumulating them and applying them all at the end.

e.g.
```
classification = ItemClassification.filler

if "trap" in item and item["trap"]:
   classification |= ItemClassification.trap # or classification = classification | ItemClassification.trap
```
etc.

FuzzyGamesOn (changes requested):

FuzzyGamesOn:
> As of recent, assigning multiple classifications to items has been added to a few of Archipelago's supported games, and it's considered to be supported behavior. Manual technically already allows multiple classifications to be assigned to an item in its .json files, but it overwrites them in sequence. This attempts to adjust the behavior so that it treats them as additive instead, in the interest of greater user freedom.

This is correct and makes sense, but is not an automatic inclusion for us. Going to mention something that we often consider for the direction of Manual:

AP supporting new behavior doesn't necessarily indicate new behavior that we want to support, because Manual prioritizes being easy to use and simple more than it does supporting AP's capabilities fully. This often comes down to weighing the value that the new behavior provides versus the complexity or confusion that it may introduce. It's fine for us to say that we don't support something if it doesn't provide enough value to our project or our users, and also fine to point the users that want it to hooks.

That being said, I can see this being useful for users. I also think keeping it out of Builder for now would avoid any confusion that newer users might have. This feels like more of a power user capability, anyways.

(Oh, and you'll want to make sure the changes account for existing uses of ItemClassification or an item's classification. DataValidation.py has an instance that would need updating, not sure if any others.)

Seatori:
> AP supporting new behavior doesn't necessarily indicate new behavior that we want to support, because Manual prioritizes being easy to use and simple more than it does supporting AP's capabilities fully. This often comes down to weighing the value that the new behavior provides versus the complexity or confusion that it may introduce. It's fine for us to say that we don't support something if it doesn't provide enough value to our project or our users, and also fine to point the users that want it to hooks.
> 
> That being said, I can see this being useful for users. I also think keeping it out of Builder for now would avoid any confusion that newer users might have. This feels like more of a power user capability, anyways.

I can respect not wanting to add every little thing. I just think that in this case, since Manual already seems so close to accepting this capability from a user standpoint, it would be a fairly harmless addition to allow users to do it. The corresponding hook method is also very simple, but that requires users to be told "you can do it using hooks" which may be an additional barrier of entry for them; one that I personally don't think is warranted given the simplicity of it.

That said, I do agree with not putting it on the Builder for now. A completely new user entering and seeing they can mix and match classifications may be an additional barrier of entry for Manual itself, especially if they're not already familiar with the concept. I believe that, as it becomes more widespread, there might eventually be a point where its inclusion would be warranted; but for now I think it makes sense as something where users can ask if it's possible and be told "yes, here's the extremely simple method of it".

> (Oh, and you'll want to make sure the changes account for existing uses of ItemClassification or an item's classification. DataValidation.py has an instance that would need updating, not sure if any others.)

Yes, I had thought of that too but I wasn't able to look into it at the time. I'll look through it while I'm looking into your other requested change!

FuzzyGamesOn:
> I did a little more testing and found that `filler_traps` adds any item flagged as a `trap` to the pool, and... apparently this behavior isn't a result of my change? It seems like it's based on what the tags are in the .json files before they're filtered out when the items are created, meaning you can already flag items as both `trap` and `useful` or `progression` and they'll always be added to the pool multiple times if filler traps are enabled.
> 
> Is this the intended behavior? If not, would this be the desired behavior in the case of this change being added?

This is a bug, yeah. The only thing that should use the raw JSON classification is create_item(), everything else should use the classification of the items in the pool by that time. In fact, that traps list and traps.append in create_items isn't needed at all; adjust_filler_items() gets a list of traps from the item pool in the else, but not in the if for some reason. So it should just do that for both.

Seatori:
> [...] In fact, that traps list and traps.append in create_items isn't needed at all; adjust_filler_items() gets a list of traps from the item pool in the else, but not in the if for some reason. So it should just do that for both (i.e., above the if else).

It looks to me like the if in `adjust_filler_items()` is supposed to be adding traps to the pool, hence the `create_item` call, while the else is for removing the traps that already exist in the pool, which is why that one searches for the items in the pool. Calling the list of items already in the pool doesn't seem like it would do much at all when it comes to adding them, and if the function was switched to only adding duplicates of the items that already exist in the pool, that would remove the current functionality of allowing certain items to only exist if filler traps are enabled.

src/__init__.py
Seatori: Forgot to leave a reply here - I changed over to your first suggestion! Thank you for the input!

FuzzyGamesOn:
> It looks to me like the if in `adjust_filler_items()` is supposed to be adding traps to the pool, hence the `create_item` call, while the else is for removing the traps that already exist in the pool...

That's a good point. Only need to make sure that your changes continue to work with what's there, then. Not super concerned about people adding prog traps or something, since they're still traps and should be treated like traps. For our purposes, it probably makes more sense for us to treat traps (and other classifications) uniformly regardless of any additional classifications on the trap. At least to start.

(Also, whenever your PR is complete, remember to hit the "Ready for review" button. Then we'll get other folks to review as well. :smiley: )

nicopop (dismissed):
this is really good!

silasary (dismissed):
Code looks good to me. I cannot test it right now, so I'll trust someone else to actually execute it :)

nicopop:
> Code looks good to me. I cannot test it right now, so I'll trust someone else to actually execute it :)

I tried it, it does work
image
the datavalidation modification will almost definetly cause a merge conflict with #104 
we'll see which get merged first the other will need a conflict fix

src/__init__.py
```diff
@@ -387,9 +386,15 @@ def adjust_filler_items(self, item_pool, traps):
             fillers = [item for item in item_pool if item.classification == ItemClassification.filler]
             traps = [item for item in item_pool if item.classification == ItemClassification.trap]
             useful = [item for item in item_pool if item.classification == ItemClassification.useful]
```
FuzzyGamesOn, line 388: Just noticed these. You'll need to find all mentions of ItemClassification and change all the `==` to `in`, or whatever's needed.

FuzzyGamesOn (changes requested):

src/__init__.py
nicopop: by using straight == it doesnt include potential prog useful thats why I didnt say anything about them

silasary: Yeah, I'm happy with these as is

FuzzyGamesOn: In a PR that adds support for multiple item classifications, it is strictly a bug to only match a single item classification with equality. Either we support multiple item classifications in all cases, or we don't. There shouldn't be any item conditionals that expect the item to not be a multiple classification item -- because with this PR, we would support users making all of their items multiple classification items. 

If the intent is to match useful that's not prog, then do useful in classification and prog not in classification. Be explicit about what should match, but make the matching support the goal of the PR.

silasary: Unless we start supporting Useful|SkipBalancing (which we don't, and also doesn't make sense), the four here (three pre-existing plus the new useful_traps) are all possible non-progression permutations of the enum

Seatori: Those are fine because they're checking for the exact values. If they were changed to `in`, they would catch items tagged with both `Progression` and `Trap`/`Useful` to be removed from the pool, which is clearly not the intent.
That's why there's a special case for `Useful + Trap` defined just below.

Seatori: Incidentally, applying `ItemClassification.filler in item.classification` and triggering this code would completely break generation, as all items contain `ItemClassification.filler` due to it being empty, resulting in it catching every single item regardless of their other flags.

Seatori: Tl;dr: The current code is correct unless we want to account for a case where a user may have manually added `Skip Balancing` to a non-progression item using hooks.

FuzzyGamesOn: Okay, I agree that the current lists will capture every permutation because filler has no value. 

However, the way that these lines of code stand currently does looks like an oversight, because checking classification equality in a post-multiple classification world seems wrong. i.e., it's not implied at a glance that useful supports useful + filler, etc. 

So I'd ask that we at least note in code comments, as part of this PR, that maintaining those lines the way they are is intended, and that it still supports all permutations of multiple classification items.

Seatori: Also, to be perfectly clear, I did search the code for instances of `ItemClassification`, and did it another time just now to be sure. The only instances I found were both evaluated and accounted for.

Seatori: GitHub only just updated to show me the other replies now. 😓 
I'll add the requested notes.

nicopop (approved):

FuzzyGamesOn (approved):
Thanks! Looks good :smiley:

silasary (approved):

PR #106, opened 2024-11-19
Add a launcher component for the Manual Discord Server

silasary:
Not sure if we want this, but it was easy enough to throw together

image

Unfortunately, I can't put it with the others, because of how/when the launcher makes the other two.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -3,6 +3,7 @@ import logging
 import os
 import json
 from typing import Callable, Optional
+import webbrowser
 
 import requests
 import Utils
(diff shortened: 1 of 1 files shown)


FuzzyGamesOn:
Hm, yeah. Normally would love to have this, but I can see the fact that it's placed above the official Discords maybe causing some drama.

They probably need to consolidate those component lists so they render all at the same time first. Then non-core stuff can add to Misc without superseding their stuff.

FuzzyGamesOn:
Does it also show up above the normal Discord buttons in the KivyMD branch? If so, I don't think we can do this until their core components are organized better to load above any client additions.

silasary:
Good question, I can't test until I'm home, but I'm assuming it'd still be above the rest. Let's just put it in adjusters and assume MD isn't merging any time soon

silasary:
image

FuzzyGamesOn (approved):
Much approve

nicopop (approved):

PR #108, opened 2024-11-21
Add search box to the client

FuzzyGamesOn:
From these feature suggestions:
https://discord.com/channels/1097532591650910289/1121577176886677506
https://discord.com/channels/1097532591650910289/1267280387877507154

This PR adds a search box to the Manual tab's view. When the contents of that search box change, the lists are immediately filtered to only results that have a match in the location or item name. Search only searches location/item names, and does not search categories or anything else. (Also, if you're wondering why showing/hiding is the way it is, it's because Kivy is unusable garbage and this was the easiest way to do it.)

This PR also opens the door for more filtering controls or similar, since any other controls can be added alongside the search layout.

---

**Screenshot of Manual client after connecting, with no search term:**

manual-client-search-empty

**Screenshot of Manual client results after entering a search term:**
_(This screenshot shows the categories expanded so it can be seen that the term matches. The categories are not auto-expanded when a search is performed.)_

manual-client-search-text

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -271,7 +271,7 @@ class ManualContext(SuperContext):
         Builder.load_string(
         """
 
-<HeaderLayout>:
+<ManualControlsStyledLayout>:
     canvas:
         Color:
             rgba: root.background_color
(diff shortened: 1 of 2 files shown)


silasary (approved):
Merge Conflicts, but I'm otherwise happy with it

silasary (approved):

nicopop (approved):

axxroytovu (approved):

PR #112, opened 2024-11-29
Make "Invalid logic format" more descriptive for requires syntax

FuzzyGamesOn:
Currently, this error is somewhat vague (should just say "requires", not "logic") and can be triggered from different areas for different reasons. Since this support question comes up decently often (twice today even, haha), wanted to clean this up.

This PR moves the KeyError up and identifies where the error came from. Then, based on that, we can offer some potential solutions to the error.

Also, better identifies whether the object passed into the error is a location or a region, and actually gives the region name if a region. (It didn't before this PR.)

(I wasn't sure if it was okay to use "enum" until I saw BaseClasses was. So using it :smile: )

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -21,7 +21,7 @@ class LogicErrorSource(IntEnum):
     EVALUATE_POSTFIX = 2 # includes missing pipes and missing value on either side of AND/OR
     EVALUATE_STACK_SIZE = 3 # includes missing curly brackets
 
-def raise_logic_error(location_or_region: dict, source: LogicErrorSource):
+def construct_logic_error(location_or_region: dict, source: LogicErrorSource) -> KeyError:
     object_type = "location/region"
     object_name = location_or_region.get("name", "Unknown")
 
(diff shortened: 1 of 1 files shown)


axxroytovu (dismissed):

silasary (changes requested):

silasary (approved):

src/Rules.py
```diff
@@ -14,9 +16,33 @@
...
+    EVALUATE_POSTFIX = 2 # includes missing pipes and missing value on either side of AND/OR
+    EVALUATE_STACK_SIZE = 3 # includes missing curly brackets
+
+def construct_logic_error(location_or_region: dict, source: LogicErrorSource) -> KeyError:
+    object_type = "location/region"
+    object_name = location_or_region.get("name", "Unknown")
+
+    if location_or_region.get("is_region", False) or "starting" in location_or_region or "connects_to" in location_or_region:
+        object_type = "region"
+    elif "region" in location_or_region or "category" in location_or_region:
```
axxroytovu, line 30: Is there a reason this isn't just an `else` statement? Both region and category are optional when defining a location so it is possible for this to break.

axxroytovu: Never mind I think I worked my way through the logic.

FuzzyGamesOn: Yeah, just coming back to this, but it's basically saying "if we can reasonably tell it's a region, it's a region; else, if we can reasonably tell it's a location, it's a location; else, default to saying it's either from above the if/else".

PR #113, opened 2024-12-02
Better messages when client errors

silasary:
Slightly improves the error messages

I currently assume that any stack trace involving UT is probably UT's fault.  I can't think of a reason that wouldn't be true.

Regardless of whether UT is involved, posts the full stack trace, not just the text parameter of the exception.

Screenshot 2024-12-02 at 11 17 25 PM

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -5,6 +5,7 @@ from typing import Any, Optional
 import typing
 from worlds import AutoWorldRegister, network_data_package
 import json
+import traceback
 
 import asyncio, re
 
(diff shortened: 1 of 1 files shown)


nicopop (changes requested):
## Important
If the window is too small the popup cover the entire area and cant be closed
image
## Optional:
the way the popup work right now does not make it easy to copy and paste the error message path.
maybe a copy to clipboard button could be helpful. 
that or a way to select the text to allow ctrl c

nicopop (commented):
1. the way the popup work right now does not make it easy to copy and paste the error message path.
maybe a copy to clipboard button could be helpful. 
that or a way to select the text to allow ctrl c
2.  when the window width is too small the errors path can go offscreen and there's no scrolling possible

silasary:
While I agree with both of these, keep in mind that all I'm doing is overriding the contents of this box:
image

I'm not quite sure how much I can push the popup without breaking AP's expectations, nor am I confident enough with kivy to do so.

nicopop (approved):
fair enough

FuzzyGamesOn (approved):

PR #114, opened 2024-12-05
After options defined use dataclass

nicopop:
As discussed on discord in the thread for pr #54 https://discord.com/channels/1097532591650910289/1311063001469812807
this make after_options_defined use the dataclass to stop dev from adding options there.

### DIFF
--- a/src/hooks/Options.py
+++ b/src/hooks/Options.py
@@ -49,24 +49,9 @@ def after_options_defined(options: PerGameCommonOptions):
     # For example if you want to change DLC_enabled's display name you would do:
     # options.__annotations__["DLC_enabled"].display_name = "New Display Name"
 
-    # The generated goal option will not keep your defined values or documentation string you'll need to add them here:
-    # To automatically convert your own goal to alias of the generated goal uncomment the lines below and replace 'Goal' with your own option of type Choice
-
-    # your_goal_class = Goal #Your Goal class here
-    # generated_goal = options.__annotations__.get('goal', {})
-    # if generated_goal and not issubclass(generated_goal, your_goal_class): #if it exist and not the exact same
-    #     values = { **your_goal_class.options, **your_goal_class.aliases } #group your option and alias to be converted
-    #     for alias, value in values.items():
-    #         generated_goal.aliases[alias] = value
-    #     generated_goal.options.update(generated_goal.aliases)  #for an alias to be valid it must also be in options
-    #
-    #     if hasattr(your_goal_class, "default"):
-    #         generated_goal.default = your_goal_class.default
-    #
-    #     if hasattr(your_goal_class, "display_name"):
-    #         generated_goal.display_name = your_goal_class.display_name
-    #
-    #     generated_goal.__doc__ = your_goal_class.__doc__ or generated_goal.__doc__
+    #  Here's an example on how to add your aliases to the generated goal
+    # options.__annotations__['goal'].aliases.update(Goal.aliases)
+    # options.__annotations__['goal'].options.update(Goal.aliases)  #for an alias to be valid it must also be in options
 
     pass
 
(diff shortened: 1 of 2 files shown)


src/hooks/Options.py
```diff
@@ -45,12 +44,16 @@ def before_options_defined(options: dict) -> dict:
...
-def after_options_defined(options: dict) -> dict:
-    # The generated goal option will not keep your defined values or documentation string unless you add them to option.json or you add them here:
-    # To automatically convert your own Goal class to alias of the generated goal uncomment the lines below and replace 'Goal' with your own option of type Choice
+def after_options_defined(options: PerGameCommonOptions):
+    # To access a modifiable version of options check the dict in options.__annotations__
+    # For example if you want to change DLC_enabled's display name you would do:
+    # options.__annotations__["DLC_enabled"].display_name = "New Display Name"
+
+    # The generated goal option will not keep your defined values or documentation string you'll need to add them here:
+    # To automatically convert your own goal to alias of the generated goal uncomment the lines below and replace 'Goal' with your own option of type Choice
```
silasary, line 53: We don't want to be encouraging people to replace Goal entirely.

nicopop: It's not replacing its converting old goal values and alias to aliases to generated goal

FuzzyGamesOn: Yeah, the examples in the comments in this might be confusing for some people. Would be good to have a clear separation between option definitions (which are their own custom ones that they make) and option modifications, which follow the same format for modifications after the fact.

So the goal example should just be `options.__annotations__.get("goal").whatever = whatever` for any of the props, instead of constructing a custom Goal option just to siphon off some properties.

FuzzyGamesOn:
Everything else looks fine other than the thing about more clear examples. Making a note that we'll (obviously) want to mention that this is a breaking change when it goes to Unstable.

FuzzyGamesOn (dismissed):
Looks good :smiley:

silasary (approved):

src/hooks/Options.py
```diff
@@ -33,39 +32,21 @@ class TotalCharactersToWinWith(Range):
...
     return options
 
 # This is called after any manual options are defined, in case you want to see what options are defined or want to modify the defined options
-def after_options_defined(options: dict) -> dict:
-    # The generated goal option will not keep your defined values or documentation string unless you add them to option.json or you add them here:
-    # To automatically convert your own Goal class to alias of the generated goal uncomment the lines below and replace 'Goal' with your own option of type Choice
+def after_options_defined(options: PerGameCommonOptions):
+    # To access a modifiable version of options check the dict in options.__annotations__
+    # For example if you want to change DLC_enabled's display name you would do:
+    # options.__annotations__["DLC_enabled"].display_name = "New Display Name"
```
silasary, line 43: This is technically incorrect.  We're passing in a `Type[PerGameCommonOptions]`, not an instance of PerGameCommonOptions.

Also, your code all works, but I'd prefer we use the the [type_hints](https://github.com/ArchipelagoMW/Archipelago/blob/main/Options.py#L1242) property over `__annotations__`.

```suggestion
def after_options_defined(options: Type[PerGameCommonOptions]):
    # To access a modifiable version of options check the dict in options.type_hints
    # For example if you want to change DLC_enabled's display name you would do:
    # options.type_hints["DLC_enabled"].display_name = "New Display Name"
```

silasary (changes requested):
Minor nitpicks

src/hooks/Options.py
nicopop: I used debugging to find a way that works. If this is better, then we use that

nicopop: Oh, your suggestion didn't include the other \__anotation__. I'll do a real commit once I'm home

FuzzyGamesOn (approved):

silasary (approved):

PR #115, opened 2024-12-10
Option to sort things by ID instead of Name

silasary:
As requested by MuffinJets

### DIFF
--- a/schemas/Manual.meta.schema.json
+++ b/schemas/Manual.meta.schema.json
@@ -86,6 +86,16 @@
             "description": "Enable the generation of puml diagram of your apworld region and locations for debug purposes",
             "type": "boolean",
             "default": false
+        },
+        "sort_items_alphabetically": {
+            "description": "Sort items alphabetically in the Client UI",
+            "type": "boolean",
+            "default": true
+        },
+        "sort_locations_alphabetically": {
+            "description": "Sort locations alphabetically in the Client UI",
+            "type": "boolean",
+            "default": true
         }
     },
     "definitions": {
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import time
 import sys
-from typing import Any, Optional
+from typing import Any, Dict, List, Optional
 import typing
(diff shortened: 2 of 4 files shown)


nicopop (approved):

FuzzyGamesOn:
imo, being able to customize sorting belongs at the client level, not at the per-apworld level as a meta setting. It should be something that people can widespread apply for all their client uses and for any Manual world from the client. More for real-time sorting by the client user, not for sorting by the world creator.

nicopop:
> imo, being able to customize sorting belongs at the client level, not at the per-apworld level as a meta setting. It should be something that people can widespread apply for all their client uses and for any Manual world from the client. More for real-time sorting by the client user, not for sorting by the world creator.

what if we let devs set a "recommended" sorting in meta.json (like this pr) and allow the players to choose to follow the devs recommendations or to override it with a setting in host.json and/or sorting buttons in client

FuzzyGamesOn:
> what if we let devs set a "recommended" sorting in meta.json (like this pr) and allow the players to choose to follow the devs recommendations or to override it with a setting in host.json and/or sorting buttons in client

That would work with a client solution. World devs can have a preference -- if they would find that useful -- but shouldn't be allowed to provide a default that overrides a player's choice or default.

---

I want to explain further so I don't come off as just preferring my own mindset or solution. The question of how we provide options for sorting comes down to the "who" and the "why" (and I think we might disagree on these):

1. To me, the "who" is the players and the "why" is "freedom to change sort order at any time while playing to find things easier". This makes it sound like a client feature.
2. This PR's solution seems to imply that the "who" is the world builders and the "why" is "because some of their world's locations don't make sense if they're not sorted to their preference". This makes it sound like a world JSON feature.

For 1, past feature requests seem to be asking for it, and it is what I assumed we were building towards.
For 2, I question why sorting would be the solution to the problem. (It's also worth noting that the same problem arises again once players have customizable sort controls if 2 is the solution.)

nicopop:
since someone recently asked/complained about the alphabetical sorting I intend to try and implement my "dev recommended sorting" idea, ill probably starts from the code of this branch and work from there and make a new PR or something

PR #116, opened 2024-12-13
remove spaces from user's options names to avoid crashes

nicopop:
Small change that make sure options added by users are compatible with python
by removing spaces in option names
Without this, spaces create some hard to understand, for non-python users, error
like this one bennydreamly got today on discord [here](https://discord.com/channels/1097532591650910289/1097891385190928504/1316918800947613696)

### DIFF
--- a/src/Helpers.py
+++ b/src/Helpers.py
@@ -68,6 +68,7 @@ def is_category_enabled(multiworld: MultiWorld, player: int, category_name: str)
 def resolve_yaml_option(multiworld: MultiWorld, player: int, data: dict) -> bool:
     if "yaml_option" in data:
         for option_name in data["yaml_option"]:
+            option_name = option_name.strip().replace(" ", "_")
             required = True
             if option_name.startswith("!"):
                 option_name = option_name[1:]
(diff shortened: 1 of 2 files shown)


src/Helpers.py
```diff
@@ -68,6 +68,7 @@ def is_category_enabled(multiworld: MultiWorld, player: int, category_name: str)
 def resolve_yaml_option(multiworld: MultiWorld, player: int, data: dict) -> bool:
     if "yaml_option" in data:
         for option_name in data["yaml_option"]:
+            option_name = option_name.strip().replace(" ", "_")
```
FuzzyGamesOn, line 71: This would be a good candidate for a helper function, since it's repeated in 5 places. Also, if we ever change how we sanitize/format option names, we'll have it in one place with a function. :smiley:

FuzzyGamesOn (changes requested):

nicopop:
I got additionnal changes to make it so the option display name keep the unformated option name (if its not already set by user) but I think ill keep it for a separate PR (so don't delete this branch on merge)

FuzzyGamesOn (approved):

silasary (approved):

PR #117, opened 2024-12-13
Finish migrating existing #about-manual docs into GitHub

FuzzyGamesOn:
Since the new documentation is in a partially written state and we have to reference both places for some answers, this PR consolidates the remaining docs in the Discord into the GitHub docs here. Following this PR, the #about-manual forums will be changed to point to these docs instead as the single source.

Few changes were made to the existing Discord docs when porting them into GitHub, and I consider improving them upon import to be outside of the scope of this PR. This PR is only to ensure that all of the docs are consolidated onto GitHub for further future refinement.

Feel free to point out anything that is glaringly wrong, though. Otherwise, just skim and go. :smiley:


nicopop (approved):

docs/design/when-youre-down.md
nicopop, line 1: this file is called "when-youre-down" instead of done

nicopop (changes requested):
just found this typo in the file name (if the readme is correct)

docs/design/when-youre-down.md
FuzzyGamesOn: Haha, good find. Will fix :smiley:

nicopop (approved):
that commit description lol

docs/faq/general.md
```diff
@@ -1 +1,67 @@
-Coming soon.
+# FAQ
+
+**Q: What is Manual?**
```
silasary, line 3: I'd rather these be `##` subheadings rather than just being bold.  

It looks better stylistically, and also gives us anchors we can link directly to

silasary (approved):
Things can definitely be improved in future PRs, but there's nothing that I'd block a merge over

docs/faq/general.md
FuzzyGamesOn: Oh yeah, lots of improvements to make over the Discord formatting. Honestly didn't review the formatting at all in GitHub because this PR is just for getting it out of Discord :smiley:

PR #118, opened 2024-12-13
Client: Fix bug that puts Victory in `(No Category)` when it has categories

FuzzyGamesOn:
There was an oversight in the way that victory categories are determined that appeared to only say that the victory location had categories if at least one of the categories it had was not shared by any other location.

This PR changes that behavior to instead use the categories that the victory location had defined. This also allows the Victory location button to show up in multiple categories now, and also prevents it from showing up in `(No Category)` when it has categories.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -520,7 +520,7 @@ class ManualContext(SuperContext):
                     if category not in self.listed_locations:
                         self.listed_locations[category] = []
                 
-                if len(victory_categories) == 0:
+                if not victory_categories:
                     victory_categories.add("(No Category)")
 
                 for category in self.listed_locations:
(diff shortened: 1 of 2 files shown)


silasary:
Oh right, we can do this now 😁

src/ManualClient.py
```diff
@@ -511,18 +511,16 @@ def build_tracker_and_locations_table(self):
...
 
-                        if category not in self.listed_locations:
-                            self.listed_locations[category] = []
-                            victory_categories.add(category)
-
-                if not victory_categories:
+                    if category not in self.listed_locations:
+                        self.listed_locations[category] = []
+                
+                if len(victory_categories) == 0:
```
silasary, line 523: Is there a reason you're using `len() == 0` instead of `not`?

FuzzyGamesOn: Stemmed from a misunderstanding of how for-else works in Python, which I originally did for the for loop above. In my head, for-else would do the for if there are any iterations and do the else if no iterations. In actuality, Python just like... does both if you don't break the loop. Which is kinda goofy, lol

I can change it to not :smile:

axxroytovu (approved):

silasary (approved):

PR #120, opened 2024-12-17
Req function argument convert rewrite

nicopop:
its the old #65 but better (hopefully)
it now can convert even weird types like: 
```python
tmp = convert_string_to_type("['test','test2']", Optional[list[str]|str|Union[int, None]] )
# tmp -> ["test","test2"]
```
also to avoid line ending issues after this I set my git to autocrlf = true and thus converted helpers back to crlf (hopefully it works this time)

I got a require function called YamlValueCompare that will use this ready.
compare with what and how you might ask? you'll see after this merge probably (don't delete the branch plz)

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -326,28 +326,28 @@ def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
 
     def convert_req_function_args(func, args: list[str], areaName: str):
         parameters = inspect.signature(func).parameters
-        knownArguments = ["world", "multiworld", "state", "player"]
+        knownParameters = ["world", "multiworld", "state", "player"]
         index = -1
-        for parameter, info in parameters.items():
-            if parameter in knownArguments:
+        for parameter in parameters.values():
+            if parameter.name in knownParameters:
                 continue
             index += 1
-            target_type = info.annotation
+            target_type = parameter.annotation
 
             if index < len(args) and args[index].strip() != "":
                 value = args[index].strip()
             else:
-                if info.default is not inspect.Parameter.empty:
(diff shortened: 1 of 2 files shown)


silasary (approved):

silasary:
Diff for Helpers.py (after ignoring whitespace changes) is:
- A couple new imports
- new function

src/Rules.py
```diff
@@ -324,74 +324,35 @@ def allRegionsAccessible(state):
...
         for parameter, info in parameters.items():
             if parameter in knownArguments:
                 continue
 
-            argType = info.annotation
-            optional = False
-            try:
-                if issubclass(argType, inspect._empty): #if not set then it wont get converted but still be checked for valid data at index
-                    argType = str
+            target_type = info.annotation
```
FuzzyGamesOn, line 335: Does this support function args that don't have a type annotation?

nicopop: That's a good question. I'll check ASAP. If it's broken here, then it's probably also broken in stable.
The function in helper.py is fine since it's asking for a type, and Python asks for a second arg, idk what gets sent when no type is provided. I'll know later today anyway

FuzzyGamesOn: Yep, no rush. :smiley: Actually just saw the removed line of `if issubclass(argType, inspect._empty):`, which seems like it was probably handling that. Inspect return annotations also do inspect._empty if there isn't one.

nicopop: You were absolutly right! On top of some broken check for type that would catch any simple types instead of just str I didnt include any check for empty. its there now

FuzzyGamesOn (approved):

silasary (approved):

PR #121, opened 2024-12-19
URL support

silasary:
This PR does two things:
* Adds support for the `archipelago://` url scheme.  This can be invoked by `Launcher.py "Manual Client" "archipelago://...?game=Manual_..."
* Updates the launcher component to register itself as a handler for any such URLs passed in without the `"Manual Client"` (This part does not currently work in 0.5.1, as the 0.5.1 version doesn't support lists of game names and only does an equality check against an assumed string)


silasary:
This doesn't actually change the client, just the launcher component.  But it is still blocked by https://github.com/ArchipelagoMW/Archipelago/pull/4133 or https://github.com/ArchipelagoMW/Archipelago/pull/3905 getting merged upstream.

FuzzyGamesOn:
Checked, those PRs are still open, so still blocked.

---

### Checked on Dec 20, 2025:
PR 4133, linked above as a blocker, is still not merged.

---

### Checked on Sep 5, 2026:
Hello 2026! 4133 still not merged :skull:

PR #122, opened 2024-12-22
Refactor Manual Core format to allow separation from data files

qwint:
adds framework to split out manual core files from the shipped manual game data format, and includes a simple way to import a ".manuworld" (theoretically can just be a .zip if that makes more sense) without hooks for a safer import

TODOs:
* flesh out the codepath where you do want to import hooks
* Options.make_options_group is completely broken and its uses commented out
* make helpers load methods work as expected and/or refactor them altogether
* figure out if one of the changes is why manual client hangs on launch and fix it
* more thorough debugging to make sure no other things need refactors

I'm not sure I'll finish this to completion, so if anyone wants to take inspiration (or the entire branch wholesale) feel free to do so
but can obviously make/approve minor changes for style/design

### DIFF
--- a/src/Data.py
+++ b/src/Data.py
@@ -1,4 +1,8 @@
 import logging
+import os
+import json
+import zipfile
+from argparse import Namespace
 
 from .DataValidation import DataValidation, ValidationError
 from .Helpers import load_data_file as helpers_load_data_file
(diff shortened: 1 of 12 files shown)


FuzzyGamesOn:
Thanks for submitting this PoC, qwint. :smiley: 

From the description, it sounds like this is an unfinished PR, so please switch it to a draft PR in the meantime. We review drafts as well, but marking it as a draft indicates that it's not expected to be ready to merge.

qwint:
yea, it's fully functional in the one codepath of load all the ".manuworld" files as json-only from the user's custom_worlds (or userdata worlds on machines that use that instead), but like the TODOs mention it does break some existing functionality so it probably shouldn't be merged as-is yet

PR #126, opened 2025-01-02
Yaml option expansion

axxroytovu:
Still needs cleanup so I'm leaving this as draft for now.

What does this PR do?

- Allows items and locations to individually have `yaml_option` definitions
- Adds syntax to allow `yaml_option` to check range & choice options:
  - `"option>10"`, `"option<10"`, `"option=10"` and `"!option=10"` all properly check against a range option
  - `"option:choice" properly checks against a choice option

### DIFF
--- a/src/Helpers.py
+++ b/src/Helpers.py
@@ -62,7 +62,8 @@ def is_category_enabled(multiworld: MultiWorld, player: int, category_name: str)
         return hook_result
 
     category_data = category_table.get(category_name, {})
-    return resolve_yaml_option(multiworld, player, category_data)
+    resolve_option = resolve_yaml_option(multiworld, player, category_data)
+    return resolve_option or resolve_option is None
 
 def resolve_yaml_option(multiworld: MultiWorld, player: int, data: dict) -> bool:
     if "yaml_option" in data:
@@ -75,7 +76,8 @@ def resolve_yaml_option(multiworld: MultiWorld, player: int, data: dict) -> bool
             option_name = format_to_valid_identifier(option_name)
             if is_option_enabled(multiworld, player, option_name) != required:
                 return False
-    return True
+        return True
+    return None
 
 def is_item_name_enabled(multiworld: MultiWorld, player: int, item_name: str) -> bool:
     """Check if an item named 'item_name' has been disabled by a yaml option."""
@@ -86,12 +88,17 @@ def is_item_name_enabled(multiworld: MultiWorld, player: int, item_name: str) ->
     return is_item_enabled(multiworld, player, item)
 
 def is_item_enabled(multiworld: MultiWorld, player: int, item: "ManualItem") -> bool:
+    from .Data import item_table
     """Check if an item has been disabled by a yaml option."""
     hook_result = before_is_item_enabled(multiworld, player, item)
     if hook_result is not None:
         return hook_result
-
-    return _is_manualobject_enabled(multiworld, player, item)
+    
+    try_resolve = resolve_yaml_option(multiworld, player, item_table.get(item.name, {}))
+    if try_resolve is None:
+        return _is_manualobject_enabled(multiworld, player, item)
+    else:
+        return try_resolve
 
 def is_location_name_enabled(multiworld: MultiWorld, player: int, location_name: str) -> bool:
     """Check if a location named 'location_name' has been disabled by a yaml option."""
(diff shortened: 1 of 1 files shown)


src/Options.py
```diff
@@ -203,6 +203,9 @@ def addOptionToGroup(option_name: str, group: str):
 
 for category in category_table:
     for option_name in category_table[category].get("yaml_option", []):
+        for c in "><=:": # Range and Choice options must be defined using Options.json
+            if c in option_name:
+                continue
```
silasary, line 208: Can you make this check for `option_name in manual_options`, and throw an exception if they forgot to define it?

silasary (changes requested):

src/Options.py
axxroytovu: Fixed in e5f27a8f6489ee0fc4211cb6f0a2a4c122c60c87

silasary (approved):

silasary (changes requested):
Choices should also use `=`, not `:`

axxroytovu:
Easy enough change.

I kept `:` instead of `=` because it felt more consistent with other manual syntax that already uses `:` but I'm not married to it.

silasary (dismissed):
If we're supporting `>=` and `<=`, we should probably also support `!=`?

axxroytovu:
It does already, but the format is "!option=value". Should be easy enough to also support "option!=value" as well.

silasary:
Yeah.  Supporting both will lead to less support requests :D

src/Helpers.py
```diff
@@ -55,27 +56,51 @@ def clamp(value, min, max):
...
 def is_category_enabled(multiworld: MultiWorld, player: int, category_name: str) -> bool:
-    from .Data import category_table
     """Check if a category has been disabled by a yaml option."""
     hook_result = before_is_category_enabled(multiworld, player, category_name)
     if hook_result is not None:
         return hook_result
 
-    category_data = category_table.get(category_name, {})
-    return resolve_yaml_option(multiworld, player, category_data)
+    category_data = multiworld.worlds[player].category_table.get(category_name, {})
```
FuzzyGamesOn, line 64: Was wondering why you referenced it this way and was reminded once again that the `is_*_enabled` hooks don't pass in world like all the other hooks, lol

src/Helpers.py
```diff
@@ -101,25 +133,28 @@ def is_location_name_enabled(multiworld: MultiWorld, player: int, location_name:
 
     return is_location_enabled(multiworld, player, location)
 
-def is_location_enabled(multiworld: MultiWorld, player: int, location: "ManualLocation") -> bool:
+def is_location_enabled(multiworld: MultiWorld, player: int, location: dict) -> bool:
```
FuzzyGamesOn, line 136: Also need to change the type hints for the hook functions named after this and the item enabled function.

src/data/items.json
FuzzyGamesOn, line 1: Need to add examples for locations.json as well since it can be used in both places.

FuzzyGamesOn (changes requested):

src/Helpers.py
```diff
@@ -55,27 +56,54 @@ def clamp(value, min, max):
...
+                option_name, target = option_name.split("!=")
+                eval_1 = lambda x, t:  x.value != t
+            elif "<" in option_name:
+                option_name, target = option_name.split("<")
+                eval_1 = lambda x, t: x.value < t
+            elif ">" in option_name:
+                option_name, target = option_name.split(">")
+                eval_1 = lambda x, t: x.value > t
+            elif "=" in option_name:
+                option_name, target = option_name.split("=")
```
nicopop, line 89: if you add == as an alternate here you'll match with what I have in #130 which im currently adding reverse so it also work like this one

src/data/items.json
axxroytovu: Done

src/Helpers.py
```diff
@@ -85,13 +116,17 @@ def is_item_name_enabled(multiworld: MultiWorld, player: int, item_name: str) ->
...
     if hook_result is not None:
         return hook_result
-
-    return _is_manualobject_enabled(multiworld, player, item)
+    
+    try_resolve = resolve_yaml_option(multiworld, player, item)
+    if try_resolve is None:
+        return _is_manualobject_enabled(multiworld, player, item)
+    else:
+        return try_resolve
```
nicopop, line 129: if I understand this part correctly if the item says its enabled via your new yaml_option then categories are not checked at all is that what you mean to do?
is that something we want?
idk but its something to think of

nicopop: actually there's a reason I originaly just put return _is_manualobject_enabled since the logic is the same for both loc and item you could move your try_resolve call there so you dont have 2 copy of this

axxroytovu: > if I understand this part correctly if the item says its enabled via your new yaml_option then categories are not checked at all is that what you mean to do? 

Hmm. I could probably put an AND in there instead

PR #130, opened 2025-01-17
Adding the YamlCompare req function

nicopop:
This Pr adds a require function that let the dev compare the yaml value of an option using any of the following:
`==`/`=` `!=` `>` `<` `>=` `<=`
and a value that get converted to wathever the compared option type support

in the format {YamlCompare(OptionName, Symbol, value to compare against)}
ex: `{YamlCompare(Example_Range > 5)}`

it works like this:
1. checks the type of the compared option
2. run the requested value in the new `convert_string_to_type` using whatever type that option supports
3. ask the option itself to try and convert the value to an int value
4. compare the converted requested value vs yaml value
    -  done by calling the comparison directly using the Comparison dict that somehow work

~~I might try and do some trickery to let user forget the `,` in the function aka have only one argument but that would require some regex that ill only create if we would prefer it that way~~ added jan 25


FuzzyGamesOn:
Seems like there's some overlap in what this and #126 are trying to solve. For consistency, feels like it would be best to have the single string argument like 126's `yaml_option` is using. Doesn't have to support that and the multiple args, either; just the single arg would work.

> have only one argument but that would require some regex that ill only create if we would prefer it that way
Shouldn't require regex. Can probably just split on operator and trim both sides, or something like that. :smile:

nicopop:
if #133 get merged before this ill add a commit to remove the unused arguments (multiworld, state, player)
otherwise if this merge before 133 ill add it in there

nicopop:
currently working on adding the `=` and !reverse from #126
I also reworked the cache so ==, !=, = saves in the same cache index

silasary (approved):

PR #131, opened 2025-01-19
Overhaul `before_create_item` hook

axxroytovu:
Nobody actually uses the current `before_create_item` hook, because it only lets you edit the item's name.

This version adds the ability to change how many of an item are created, which is a significantly more stable way to manage dynamic item counts than the current process of removing them in `before_create_items_filler`

The old functionality is still maintained

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -131,7 +131,7 @@ class ManualWorld(World):
 
             if item_count == 0: continue
 
-            name, item_count = before_create_item(name, item_count, self, self.multiworld, self.player)
+            name, count = before_create_item(name, count, self, self.multiworld, self.player)
 
             for _ in range(item_count):
                 new_item = self.create_item(name)
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -131,8 +131,6 @@ class ManualWorld(World):
 
             if item_count == 0: continue
 
-            name, count = before_create_item(name, count, self, self.multiworld, self.player)
-
             for _ in range(item_count):
                 new_item = self.create_item(name)
                 pool.append(new_item)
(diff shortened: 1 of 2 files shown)


FuzzyGamesOn:
The `before_create_item` hook is intended for cases where a user might want to intercept a create item request for one item but instead create another item based on some logic. e.g., if their options could cause one item to be entirely swapped out for another. Also lets them intercept create items that are triggered from other places than `create_items`, like a YAML start_inventory. 

I do like the idea of adding a simple hooks way to modify count, though. Have felt like `create_items` needed an earlier hook that lets the user customize some parts of the item record before we start processing it, and I agree that the current way of modifying the item pool is a bit convoluted compared to pre-editing the item in the loop.

Could we instead do this functionality as a new hook? Name can be whatever as long as it starts with `before_create_items_` for consistency with the other hooks in `create_items`. Would also consider if there are any non-name properties that we'd want to allow the user to edit in addition to count.

axxroytovu:
Completely overhauled this PR.

Now the `before_create_items_all` hook allows you to access a new dictionary that contains information about how the item should be generated. That dictionary can have the following formats:
```py
{"Item Name": 5} # <- This will create qty 5 items using all the default settings
{"Item Name": {"useful": 7}} # <- This will create qty 7 items and force them to be classified as useful
{"Item Name": {"progression": 2, "useful": 1}} # <- This will create 3 items, with 2 classified as progression and 1 as useful
{"Item Name": {0b0110: 5}} # <- If you know the special flag for the item classes, you can also define non-standard options. This setup
#       will create 5 items that are the "useful trap" class
```

silasary (approved):

FuzzyGamesOn (approved):

PR #132, opened 2025-01-19
Add .editorconfig to standardize encoding and whitespace

silasary:
Similar to #129, this standardizes a few IDE settings to reduce unintended whitespace and encoding changes

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.{py,md,json}]
+indent_style = space
+indent_size = 4
+charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.{py,md,json}]
```
nicopop, line 6: .md documentation can have 2 trailing whitespace (to force a new line)
So you'll need to split them from this tag

nicopop (changes requested):
the 1 issue I found while trying to write docs for something with this config included

FuzzyGamesOn (approved):

nicopop (approved):

PR #133, opened 2025-01-21
Make req function generic arguments optional

nicopop:
this let devs make their require functions arguments in anyway they wants.
if they want to have the world all they need to do is include an argument with any name with `: World`
same for CollectionState and Multiworld.
the only reserved argument name is player
example:
```python
SurpriseMe(wld: World, mv: Multiworld, player: int)
```
is valid
as is
```python
CopyPastaRequire()
```
Even the following would work:
```python
scrambledArgs(test: str, world: World, player: int, potato: str)
```
I also removed the unused arguments of existing req rules to fit with this change.

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -423,7 +423,7 @@ def ItemValue(world: World, multiworld: MultiWorld, state: CollectionState, play
     return world.itemvalue_rule_cache[player][value_name]['count'] >= requested_count
 
 # Two useful functions to make require work if an item is disabled instead of making it inaccessible
-def OptOne(world: World, multiworld: MultiWorld, state: CollectionState, player: int, item: str, items_counts: Optional[dict] = None):
+def OptOne(world: World, item: str, items_counts: Optional[dict] = None):
     """Check if the passed item (with or without ||) is enabled, then this returns |item:count|
     where count is clamped to the maximum number of said item in the itempool.\n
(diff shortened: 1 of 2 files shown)


src/Rules.py
```diff
@@ -335,22 +335,32 @@ def allRegionsAccessible(state):
...
+            if target_type in knownParameters or parameter.name.lower() == "player":
+                if target_type in [World, 'ManualWorld']:
+                    args.insert(index, world)
+                elif target_type == MultiWorld:
+                    args.insert(index, multiworld)
+                elif target_type == CollectionState:
+                    args.insert(index, state)
+                else:
+                    args.insert(index, player)
+                continue
```
silasary, line 354: Slightly easier to read if the player check is separate
```suggestion
            if target_type in knownParameters:
                if target_type in [World, 'ManualWorld']:
                    args.insert(index, world)
                elif target_type == MultiWorld:
                    args.insert(index, multiworld)
                elif target_type == CollectionState:
                    args.insert(index, state)
                continue
            if parameter.name.lower() == "player":
                args.insert(index, player)
                continue
```

silasary (changes requested):
Minor nitpick, but otherwise good

silasary (approved):

FuzzyGamesOn (approved):

PR #135, opened 2025-01-22
uses explicit tab adding to not get the needless "All" tab

qwint:
to keep back compat with before 0.5.1 you can reimplement add_client_tab() https://github.com/ArchipelagoMW/Archipelago/pull/3950
but that function was created so clients wouldn't have to change when kivymd gets merged so I would suggest using it for future compat

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -14,9 +14,6 @@ ModuleUpdate.update()
 
 import Utils
 
-if __name__ == "__main__":
-    Utils.init_logging("ManualClient", exception_logger="Client")
-
 from NetUtils import ClientStatus
 from CommonClient import gui_enabled, logger, get_base_parser, ClientCommandProcessor, server_loop
 from MultiServer import mark_raw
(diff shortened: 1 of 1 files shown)


qwint:
Only thing it should do from a UI perspective is remove the "All" tab that duplicates the "Archipelago" tab (but obviously not stopping you from doing testing haha)

silasary:
While probably not important, it's worth noting that this does move the tab to the end
image

silasary (approved):

FuzzyGamesOn:
Ah, yeah. I figured it might do that. Have a vague memory of trying that while working on other client changes.

Yeah, to me, it's actually more important to have the Manual tab (the one that users will use the most, other than the AP tab) near the front for usability reasons. People ignore the "All" tab, anyways. If `add_client_tab()` or Kivy tabs in general supported adding a tab at a certain position, we could do that and this change would be a no-brainer... but when I glanced at these prior, it didn't seem like this worked or was available. Maybe this is another thing that KivyMD will fix over Kivy, no idea.

I'm also not convinced that this alone will make our client compatible with the KivyMD change, so I don't feel like we're at the point of compromising the above yet.

FuzzyGamesOn:
The fact that this subtle tab change takes our CommonClient-extended client from completely broken in KivyMD to functional in KivyMD... really speaks to how much of a complete fuckery the KivyMD update in AP is.

qwint:
> While probably not important, it's worth noting that this does move the tab to the end

FYI https://github.com/ArchipelagoMW/Archipelago/pull/4876 adds an index arg to add_client_tab so it should be easy to reorient the tabs as you wish now

PR #138, opened 2025-02-02
Add component to launcher even if Launcher thinks it's called __main__

silasary:
I made an oversight in  #124, this fixes it


nicopop (dismissed):
Looks logical, aka I've never heard of that issue, but I trust you that it exists, and this looks like the way to fix it

FuzzyGamesOn (approved):

nicopop (approved):

PR #139, opened 2025-02-20
Dev recommended item/loc sorting

nicopop:
Alternate continuation of #115 
it didnt feel polite to continue on their branch/pr so I made mine and merged from theirs
if silasary wants to continue their own version of this then I can close this PR

Even if AP ClientSide is something I barely know I can work with what's already there or get inspired from other apworlds if required (minecraft fabric client was a big source of info on how to use and save settings)
## Anyway
As discusted, in #115 I let devs set their prefered sorting out of [Id, inverted_id, alphabetical, inverted_alphabetical]
why an enum and not a bool? its so later we can add more sorting methodes/algorithm
Of course I added some settings so the player can override the dev's sorting preference.
the settings get saved in the host.yaml

the one thing im not sure about that is the setting key, or in the host.yaml should every manual's setting use a single "manual_settings" key of have one for each manual
(currently going for singular but im leaving the code commented out for multiple)

Also I dont know how to make buttons so commands is what you'll get for now
If anyone wants to participate in the PR to add those or find a better way to refresh the locations on change of sorting then feel free to do so

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -2,8 +2,9 @@ from base64 import b64encode
 import logging
 import os
 import json
-from typing import Callable, Optional
+from typing import Callable, Optional, Union
 import webbrowser
+import settings
 
 import Utils
 from worlds.generic.Rules import forbid_items_for_player
@@ -11,7 +12,7 @@ from worlds.LauncherComponents import Component, SuffixIdentifier, components, T
 
 from .Data import item_table, location_table, region_table, category_table
 from .Game import game_name, filler_item_name, starting_items
-from .Meta import world_description, world_webworld, enable_region_diagram, sort_items_alphabetically, sort_locations_alphabetically
+from .Meta import world_description, world_webworld, enable_region_diagram, preferred_items_sorting, preferred_locations_sorting, SortingOrder
 from .Locations import location_id_to_name, location_name_to_id, location_name_to_location, location_name_groups, victory_names
 from .Items import item_id_to_name, item_name_to_id, item_name_to_item, item_name_groups
 from .DataValidation import runGenerationDataValidation, runPreFillDataValidation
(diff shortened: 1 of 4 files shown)


FuzzyGamesOn:
I feel like this PR is missing the point of what was discussed in #115. To be completely honest, a sort preference that the world dev sets is the lowest importance thing that was discussed in that PR. 

#115 didn't progress further because it left the default sort order up to the world dev, which fails to give Manual players a consistent, world-independent default sort order, which is what the current alphabetical sorting provides. It doesn't matter *what* the default sort order is, but it has to be consistent across all worlds as a default. When players ask "how are these sorted by default", our answer should be clear. 

So, if players are unhappy with the default sort order in the unstable, then we have to change the default sort order. Then we can add on as many sort modifiers as we want after. But first step is "change alphabetical to something else". Axx suggested doing it by id instead in https://discord.com/channels/1097532591650910289/1097891385190928504/1315877764083220500, and I think that's a viable idea. :smiley:

silasary:
I've only glanced at the code briefly, but my impression is that the author defined sort order only applies if the user configures it to "recommended". 

The moment they preference always sorting by name/id, the apworld preference goes out the window.

(If I'm wrong and that's not the behaviour, it probably should be)

nicopop:
@FuzzyGamesOn 2 things,
1. I need to redo the description I was tired and I should've went with the draft pr but I had the requested review already selected and didnt know if it would still ping you if it was a draft and at the time it felt done enough to still press send
2. This pr let devs set a recommended sorting yes, but it let the user override the default by doing /items/locations_sorting to anything other than recommended and that saves to hosts.yaml so next time the user loads a manual it use their saved setting

nicopop:
also if we want preference to go back to id then I literaly need to change the code in I think 3 lines

silasary:
I think in absence or user or author preferences, we should do Alphabetical by default.  It's usually a better experience, and is what the current Unstable is doing.

silasary:
oooh, I like those

src/__init__.py
FuzzyGamesOn, line 1: Need to bump client version at the bottom of init when updating the ManualClient.

FuzzyGamesOn (changes requested):

FuzzyGamesOn (approved):

nicopop:
Ill say this PR is ready, if/when I want to re-add the dev recommended part it will be in another PR

silasary (dismissed):

silasary (dismissed):

silasary (approved):

PR #140, opened 2025-02-24
Fix region logic to apply to entrances and not exits

axxroytovu:
Region logic was applied to exits, and so `can_reach_region` was incorrectly triggering before the region requirements were found.


silasary (approved):

silasary:
Ah, yep.  I see the bug there

FuzzyGamesOn (approved):

PR #141, opened 2025-03-08
Implement Events

axxroytovu:


axxroytovu:
Ok, we're very close. Events now have an optional "visible" tag that will make all copies of that event visible in the client. Still to-do:

- [ ] More rigorous testing
- [ ] Add syntax example for visible
- [ ] Add schema

nicopop (approved):
Im aproving without testing it since silasary said they would test, that and the code look good

nicopop (dismissed):

nicopop (approved):

silasary (dismissed):
Let's merge this 😀

nicopop (approved):
after that little schema fix this should work

PR #142, opened 2025-03-13
User facing events

silasary:
NOTE:  Very work-in-progress.

This PR lets authors make pseudo-locations and place events on them.  These are useful for logic reasons (They can be required by the generator, and should cut down on CanReach usage), but with the aid of Universal Tracker, can also place tracker items into the player's inventory, allowing for simple logic calculations without adding place_item checks that inflate the location count.

It is not done yet, validation rules will fail on non-item event requirements, and I haven't tested the UI anywhere near enough.   But pushing for the sake of sharing where it's at, since I probably won't get back to it for another week or so


silasary:
Closing this in favour of #141

PR #144, opened 2025-04-17
Client sort order revert while we wait on client updates

FuzzyGamesOn:
Sort order is a blocker for the next unstable, and we've paused client updates in the meantime, and we already have to accept some concession because of the KivyMD changes... so this PR is an additional concession to change sort order to id / non-alphabetical in the meantime. 

Goal is to revisit sort order once client changes are finalized to some extent.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -521,7 +521,7 @@ class ManualContext(SuperContext):
                     victory_categories.add("(No Category)")
 
                 for category in self.listed_locations:
-                    self.listed_locations[category].sort(key=self.ctx.location_names.lookup_in_game)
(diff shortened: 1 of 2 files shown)


silasary (approved):
I maintain that ID sorting is an anti-pattern and something we should be discouraging. 

But we need to get a stable release out and the good solutions aren't ready.

nicopop (approved):
I somewhat agree with Silasary and this will do for now

FuzzyGamesOn:
> I maintain that ID sorting is an anti-pattern and something we should be discouraging.
> 
> But we need to get a stable release out and the good solutions aren't ready.

Yep, I 100% agree with all of this as well. :smile:

PR #145, opened 2025-04-21
fix crash cause by unbound item in create_items

nicopop:
Found this while redoing ItemValue to work more like AP does things:
this small change from [#131](https://github.com/ManualForArchipelago/Manual/pull/131/files#diff-6b39a2f1a6399aa39e427fb4c11a7b552e615a2d83a3c36aa6dd6ddbbb8281f4R120-L127) cause a crash


axxroytovu (approved):
Good catch!

silasary (approved):

PR #146, opened 2025-04-21
ItemValue Simplified

nicopop:
Last december qwint shared a better way to do item values [here](https://discord.com/channels/1097532591650910289/1097538232914296944/1314704842731815004)
this is the implementation of it.

I also replaced all the wrong `any` to `Any` in helpers.py while I was there

This works by making it that every item when collected, add their values to the state that acts like regular items that can be found with state.has/count no more silly cache.

Since this uses regular AP logic, it become more open to outside modifications. 
I added Hooks so devs can do something whenever collect/remove is called.
It could let devs change the value of items or even create progressive items

### DIFF
--- a/src/Helpers.py
+++ b/src/Helpers.py
@@ -4,6 +4,7 @@ import pkgutil
 import json
 
 from BaseClasses import MultiWorld, Item
+from enum import IntEnum
 from typing import Optional, List, TYPE_CHECKING, Union, get_args, get_origin, Any
 from types import GenericAlias
 from worlds.AutoWorld import World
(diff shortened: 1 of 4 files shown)


nicopop:
1am idea from my bed, I need to note somewhere: 
* Do the same for categories (add them in collect/remove)
* convert the get key helper func to input category of key (category instead of just VALUE)

nicopop:
last night me, you have good idea but probably not for this PR, ill add some hooks for create/remove and this pr should be done, more can be done later

nicopop:
this PR is done for now, more can be added later PRs

FuzzyGamesOn (dismissed):

FuzzyGamesOn (approved):

silasary (approved):

PR #148, opened 2025-05-01
Fix before_create_items_all hook

nicopop:
for a manual im doing I decided to use the before_create_items_all hook, but using any dict would fail 
After fixing that I decided to test everything the comments says should work and thus found that the binary version mentionned in the comments also do not work

Here is what the comments looks like for the hooks currently:
```python
# This hook allows you to access the item names & counts before the items are created. Use this to increase/decrease the amount of a specific item in the pool
# Valid item_config key/values:
# {"Item Name": 5} <- This will create qty 5 items using all the default settings
# {"Item Name": {"useful": 7}} <- This will create qty 7 items and force them to be classified as useful
# {"Item Name": {"progression": 2, "useful": 1}} <- This will create 3 items, with 2 classified as progression and 1 as useful
# {"Item Name": {0b0110: 5}} <- If you know the special flag for the item classes, you can also define non-standard options. This setup
#       will create 5 items that are the "useful trap" class
```
before 9fb61d1 only the first example worked
before 3896093 the last example wouldnt work

I decided to change from the predefined dict of ItemClassification to just calling ItemClassification itself since its simpler and is more future proof if base AP add a new classification later or something
and while at it I just let devs provide the ItemClassification directly in a2b66dd

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -132,13 +132,13 @@ class ManualWorld(World):
 
         for name, configs in items_config.items():
             total_created = 0
-            if type(configs) == int:
+            if type(configs) is int:
                 total_created = configs
                 for _ in range(configs):
                     new_item = self.create_item(name)
                     pool.append(new_item)
-            elif type(configs) == dict:
-                for cat, count in configs.values():
+            elif type(configs) is dict:
+                for cat, count in configs.items():
                     total_created += count
                     true_class = {
                         "filler": ItemClassification.filler,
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -132,23 +132,25 @@ def create_items(self):
...
-                        "trap": ItemClassification.trap,
-                        "useful": ItemClassification.useful,
-                        "progression_skip_balancing": ItemClassification.progression_skip_balancing,
-                        "progression": ItemClassification.progression
-                    }.get(cat, cat)
-                    if not isinstance(true_class, ItemClassification):
-                        raise Exception(f"Item override for {name} improperly defined")
+                    if isinstance(cat, ItemClassification):
+                        true_class = cat
+                    else:
```
axxroytovu, line 145: I did have to fix this for proper integer handling:
```
                    else:
                        try:
                            if isinstance(cat, int):
                                true_class = ItemClassification(cat)
                            elif cat.startswith('0b'):
                                true_class = ItemClassification(int(cat, base=0))
                            else:
                                true_class = ItemClassification[cat]
```

axxroytovu (changes requested):

src/__init__.py
nicopop: done.

axxroytovu (approved):

PR #149, opened 2025-05-06
Ensure json rules cannot run python code while being parsed

Mysteryem:
Hooks and modifications to the manual world itself are expected places that could run any python code. What is not expected is json rules themselves running python code when being parsed.

Due to the use of `eval()`, function arguments within json rules could have run arbitrary code.

The use of `eval` has been replaced with `ast.literal_eval`, ensuring that function arguments being parsed in json rules cannot run arbitrary Python code.

---

Json rules should be considered 'safe', not requiring any sort of review, in my opinion. Without this change, json rules are currently 'unsafe' and would have to be reviewed alongside any hooks a custom manual world adds, or any changes a custom manual world makes to the base manual world.

As a note, the code that is using `eval` is not present in any stable or unstable releases of Manual yet.

### DIFF
--- a/src/Helpers.py
+++ b/src/Helpers.py
@@ -1,3 +1,4 @@
+import ast
 import csv
 import os
 import pkgutil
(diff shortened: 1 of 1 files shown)


silasary (approved):

silasary:
Good catch :)

FuzzyGamesOn (approved):

PR #150, opened 2025-05-12
fix the item counts for location requirements desync with UT

nicopop:
Seems the item_counts was made too late and for some reason UT and only UT couldn't see the precollected items
by creating it at the start of set_rules UT can now see precollected items too

this makes it work with both game.json's starting_items and the player yamls's start_inventory and start_inventory_from_pool
in start_inventory case item are added to the pool and just added to the counts thus for  :all and any :% will grow with the items
but start_inventory_from_pool should now work correctly since the items already exists

the only way this will break is if a custom rule made in the rules.py hook call the reset of the item count cache and thus break UT again
Im not sure the best way to prevent that from happening, could potential create an invalid player in the cache (-1) and test for it and not reset if present or something

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -102,6 +102,7 @@ def evaluate_postfix(expr: str, location: str) -> bool:
     return stack.pop()
 
 def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
+    world.get_item_counts(player, True)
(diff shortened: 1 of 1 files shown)


FuzzyGamesOn (approved):
Seems like a good addition :smile:

axxroytovu (approved):

silasary (approved):

PR #154, opened 2025-05-31
Fix syntax error with f-string

silasary:
I'm not sure how this snuck in, but it's definitely subtle enough that I'm not surprised


nicopop (approved):
wow,
I probably edited/moved some code wrong and that slipped in or something 😮

axxroytovu (approved):

FuzzyGamesOn (approved):

PR #155, opened 2025-06-06
ci: Add Eijebong's testing github actions

silasary:
Tests 🎉

### DIFF
--- a/.github/workflows/fuzz.yaml
+++ b/.github/workflows/fuzz.yaml
@@ -1,0 +1,16 @@
+name: Fuzz apworld
+on:
+  push:
+  pull_request:
+
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Eijebong/ap-actions/fuzz@main
+        with:
+          apworld-path: src
+          ap-version: '0.6.1'
+          python-version: '3.12'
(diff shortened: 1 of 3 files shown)


.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,16 @@
...
+  pull_request:
+
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Eijebong/ap-actions/fuzz@main
+        with:
+          apworld-path: src
+          ap-version: '0.6.1'
```
FuzzyGamesOn, line 13: Instead of hard-coding AP version, we can get it from their GitHub latest release in a previous step. Some random Bash that will get the version number from the URL:

`wget https://github.com/ArchipelagoMW/Archipelago/releases/latest 2>&1 | grep Location: | awk '{print $2}' | rev | cut -d "/" -f 1 | rev`

silasary: Just doing it with the bash script for now.  We can update it once Eij implements 'latest' support

nicopop (approved):
looks logical based on my limited knowledge of github workflows

FuzzyGamesOn (approved):

PR #156, opened 2025-06-12
Simplifying Item_counts

nicopop:
Rework get_item_counts to only create the item counts once(1 time) at the end of create_items
this should once and for all fix the counts by using the pool straight out of create_items instead of trying to recreate the item pool later from multiworld and precollected
I also made rules get the count of progression items by default ~~but
I also included a ! suffix if people want to use the old logic of All being every copy of an item not just the ones with progressions in their classification 
the ! suffix is a something that can and probably should be something else is I just dont know what, It could even be removed entirely~~ since I remembered State doesnt count non-progression items while generating meaning any All! would potentially never be in logic while generating

This was tested  in 2 tests using the same test manual with 3 copies of an item with 1 in game.json starting_items
- [x] 1.  1 in start_inventory_from_pool
- [x] 2.  1 in start_inventory

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -58,7 +58,8 @@ class ManualWorld(World):
 
     filler_item_name = filler_item_name
 
-    item_counts = {}
+    item_counts: dict[int, dict[str, int]] = {}
+    progression_counts: dict[int, dict[str, int]] = {}
     start_inventory = {}
 
     location_id_to_name = location_id_to_name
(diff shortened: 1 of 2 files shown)


src/__init__.py
```diff
@@ -58,7 +58,8 @@ class ManualWorld(World):
 
     filler_item_name = filler_item_name
 
-    item_counts = {}
+    item_counts: dict[int, Counter[str]] = {}
+    progression_counts: dict[int, Counter[str]] = {}
```
FuzzyGamesOn, line 62: As part of this PR, it looks like all calls to `world.get_item_counts()` were changed to do only progression, and I don't see any other references to world.item_counts that use it in a non-prog way. (And pseudo-keywords only do state, so only prog.) So I have two questions:

1. What's the use case for having a non-prog `item counts`? What uses it or could use it?
2. If prog is the default usage, shouldn't `get_item_counts()` default to prog_only?

FuzzyGamesOn: I guess there's some potential value to being able to get overall counts at the end of `create_items`, even if we don't use it. Thanks for clarifying. :smile:

FuzzyGamesOn (approved):

silasary (approved):
I'm happy with this

PR #167, opened 2025-09-02
Count hook precollected items too

nicopop:
As the name suggest this PR make sure Manual count precollected items from hooks in create_items
and thus work with :all, :half and others dynamic requirements


axxroytovu (dismissed):
Seems simple enough

PR #201, opened 2026-02-02
Made Events work with @requirement and RB bugfixes

nicopop:
I got inspired by what I said on discord in manual-core-dev channel and I went and really looked at the rules.py file
things included in this PR so far:
- Replaced the old state.counts call loops with calls to either state.has or state.has_from_list
  - fixes events not being accessible via `@category`, `:half/:all`
- Added Some type anotations to make mypy happy
- Made item_count convert back to item_name if the count is an unknown string

that last point can be reverted/modified if we still want it to raise an exception instead

### DIFF
--- a/src/Rules.py
+++ b/src/Rules.py
@@ -45,10 +45,10 @@ def construct_logic_error(location_or_region: dict, source: LogicErrorSource) ->
 
     return KeyError(f"Invalid 'requires' for {object_type} '{object_name}': {source_text} (ERROR {source})")
 
-def infix_to_postfix(expr, location):
-    prec = {"&": 2, "|": 2, "!": 3}
-    stack = []
-    postfix = ""
+def infix_to_postfix(expr: str, location: dict) -> str:
+    prec: dict[str, int] = {"&": 2, "|": 2, "!": 3}
+    stack: list[str] = []
+    postfix: str = ""
 
     try:
         for c in expr:
@@ -73,8 +73,8 @@ def infix_to_postfix(expr, location):
     return postfix
 
 
-def evaluate_postfix(expr: str, location: str) -> bool:
-    stack = []
+def evaluate_postfix(expr: str, location: dict) -> bool:
+    stack: list[bool] = []
 
     try:
         for c in expr:
@@ -167,9 +167,9 @@ def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
                 require_category = True
 
             item_base = item
-            item = item.lstrip('|@$').rstrip('|')
+            item: str = item.lstrip('|@$').rstrip('|')
 
-            item_parts = item.split(":")  # type: list[str]
+            item_parts: list[str] = item.rsplit(":", 1)
             item_name = item
             item_count: str | int = "1"
 
@@ -338,7 +338,7 @@ def set_rules(world: "ManualWorld", multiworld: MultiWorld, player: int):
     # Victory requirement
     multiworld.completion_condition[player] = lambda state: state.has("__Victory__", player)
 
-    def convert_req_function_args(state: CollectionState, func, args: list[str], areaName: str):
+    def convert_req_function_args(state: CollectionState, func, args: list[str| Any], areaName: str):
         parameters = inspect.signature(func).parameters
         knownParameters = [World, 'ManualWorld', MultiWorld, CollectionState]
         index = -1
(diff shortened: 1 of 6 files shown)


silasary (dismissed):

src/Rules.py
```diff
@@ -426,7 +424,10 @@ def fullLocationOrRegionCheck(state: CollectionState, area: dict):
         used_location_names.extend([l.name for l in multiworld.get_region(region, player).locations])
         for exitRegion in multiworld.get_region(region, player).entrances:
             extra = extra_entrance_rules.get(exitRegion.name, {})
-            rb_rule = construct_rule_from_string(regionMap[region])
+            area = regionMap[region]
```
silasary, line 427: If we're reassigning ["name"], this should probably be a `.copy()`

silasary (approved):

src/Rules.py
nicopop: its what we did before RB, I just moved it here so errors after this part of the code have a name and are detected as a region.
one thing we could do instead is move that code to be in regions.py so that they use the exact region name instead of `region A to B` thing they have currently (if i remember correctly)

silasary: Yeah, I'm aware that it's something we're already doing, that's why I approved regardless. 

But when you moved it it brought it to my attention as a thing we probably shouldn't be doing.

src/Rules.py
```diff
@@ -163,10 +163,10 @@ def recursively_tokenize_manual_rule(partial: str) -> "rule_builder.rules.Rule |
                 if item_count.isnumeric():
                     count = int(item_count)
                 else:
-                    count = evaluate_nonnumeric_count(item_name, item_count, is_category, area)
+                    item_name, count = evaluate_nonnumeric_count(match.group(0), item_name, item_count, is_category, area)
 
                 if is_category:
-                    rule = rule_builder.rules.HasGroup(item_name, count)
+                    rule = rule_builder.rules.HasFromList(*world.item_and_event_name_groups.get(item_name, set()), count=count)
```
silasary, line 169: Wait, why are we changing this?

nicopop: To include events since those cant be on item groups according to AP tests

silasary: Oh, I see.  Yeah, makes sense.

nicopop:
Found out recently while doing a Manual with all of my PRs that with RB there are no call to construct_logic_error for unbalanced parentheses (which causes an exception in the code) idk If I should include the fix for it here or in another PR after this one

nicopop:
> Found out recently while doing a Manual with all of my PRs that with RB there are no call to construct_logic_error for unbalanced parentheses (which causes an exception in the code) idk If I should include the fix for it here or in another PR after this one

The reason I wonder if I should do it in another PR is that while fixing this I found other issues with the way Manual react when there's an invalid item/function with a number in it
the fix for that bug could be in a following PR or it could be in this one

---

A require of `|Chun-Li| or {YamlCompare(Example_Choice == 1)` doesnt trigger any exception
it goes in the following order
1. RB fails because it cannot be tokenized
2. manual fallback to checkRequireStringForArea
3. it also fail at detecting the "function" since its invalid
4. infix_to_postfix does it thing skip all the non-numeric character
5. it resolve to 0|1 because of the `== 1`
6. manual see that as valid and voila a missed typo

the fix for this I have locally is to make `infix_to_postfix` raise a valueError on non-numeric char 
and then construct a more "correct" logic error
it does mean the exception checking of `evaluate_postfix` would probably become useless since they would now be done in infix_to_postfix and that we probably could modify the LogicErrorSource values name

```python
def infix_to_postfix(expr: str, location: dict) -> str:
            ...
            elif c == ")":
                while stack and stack[-1] != "(":
                    postfix += stack.pop()
                stack.pop()
            else:
                # added this here since '|Chun-Li| or {YamlCompare(Example_Choice == 1)' is valid because of the 1 at the end
                raise ValueError(f"Invalid Character '{c}' in expression '{expr}', it should be either a number or a parentheses")

        while stack:
            postfix += stack.pop()
    except ValueError as ex:
        text = str(ex)
        if "'{'" in text or ")}" in text:
            raise construct_logic_error(location, LogicErrorSource.EVALUATE_STACK_SIZE)

        raise construct_logic_error(location, LogicErrorSource.EVALUATE_POSTFIX)
    except Exception:
        ...
```

FuzzyGamesOn (approved):

nicopop:
I merged this knownly that ill need to make another PR to fix my last 2 comments

PR #205, opened 2026-02-18
Feature: location scouting

Et0san:
Adds a scout button to locations set as scoutable: true in locations.json, that sends a hinted LocationScout.
I'm a newbie to Kivy, so I kept the edits as light as I could, but this could take a revamp of the location display structure to show better, with a fancy icon instead of the "Scout" text, both buttons in a proper BoxLayout so that we don't need placeholders in a 2 cols GridLayout and so that non-scoutable locations take the same space as scoutable (400px vs 350+50px, or 370+30 if we manage the fancy icon).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 .vs/
 /*.code-workspace
 .idea
+build.py
+*.apworld
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -1093,7 +1093,13 @@ class ManualContext(SuperContext):
                                             category_count += 1
 
                                 for location_button in buttons_to_remove:
-                                    location_button.parent.remove_widget(location_button)
+                                    parent = location_button.parent
+                                    button_index = parent.children.index(location_button)
+                                    # Grid cols=2: pairs are at adjacent indices (even with odd, odd with even)
+                                    pair_index = button_index - 1 if button_index % 2 == 1 else button_index + 1
+                                    if 0 <= pair_index < len(parent.children):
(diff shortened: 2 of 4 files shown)


silasary (changes requested):
Gave this a quick spin, haven't done a full code review yet. 

But two immediate things I noticed that should be changed:

* You need to bump the client version number in `__init__.py`.   This is something I really need to document better 😅

* I don't like that the Scout button is still visible and unchanged once you've scouted a location. 
I can a few possible solutions:
  * Replace "Scout" with "Scouted"
  * Remove button entirely 
  * Replace button with Label saying "contains Player1's Item"

My preference is the third one - applied to all hints (not just scouts) - but it is the most work of the three.

silasary (dismissed):
It's in a good state now :)

nicopop (dismissed):
~~I hope this doesnt break too much of my #214 but~~ otherwise it looks good
I hope this (205) get merged soon so I can deal with the merge conflict on my PR ;)
edit: did a test merge of every Client PR we have as of April 13 and it had no major conflict, just small tweaks

silasary:
@Et0san can I get your eyes on the changes I made to this PR?

Et0san:
Looks good! I don't know if it's feasible to prevent the UI from fully refreshing (and closing our opened categories; or if refreshing remembering the categories' deployed states)?

silasary:
Rewriting the entire thing to use a recycleview is absolutely something we can and probably should do. But it needs to be a standalone PR with a lot of testing

silasary (approved):

nicopop (approved):

PR #209, opened 2026-03-20
Add UT glitch logic support

nicopop:
This PR add support for UT OOL location display.
This is done using a special Event Item that can be added to any requirement.
Said item's name can be set in game.json.
Any location which requirement work with this item get displayed in Yellow in the tracker tab and with this PR in our tab too
With this PR any category with a location available OOL will have the count of OOL location added next to its original count in parenteses
image

I added new settings to disable this OOL display entirerly and choose if this bypasses the location button protection
image

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -477,7 +477,7 @@ class ManualContext(SuperContext):
                 config.setdefaults("universal-tracker", {
                     "block_unreachable_location_press": "Yes",
                     "display_glitched_locations": "Yes",
-                    "allow_glitched_location_press": "No"
+                    "allow_glitched_location_press": "Yes"
                 })
 
             def build_settings(self, settings: Settings):
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -121,7 +121,8 @@ class ManualContext(SuperContext):
     region_table = {}
     category_table = {}
 
-    tracker_reachable_locations = []
+    tracker_reachable_locations: list[str] = []
+    tracker_glitched_locations: list[str] = []
     tracker_reachable_events = []
 
     set_deathlink = False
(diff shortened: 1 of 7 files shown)


src/ManualClient.py
```diff
@@ -453,7 +475,9 @@ def build_config(self, config: ConfigParser):
                     "locations_sorting_order": SortingOrderLoc.default.name
                 })
                 config.setdefaults("universal-tracker", {
-                    "block_unreachable_location_press": "Yes"
+                    "block_unreachable_location_press": "Yes",
+                    "display_glitched_locations": "Yes",
+                    "allow_glitched_location_press": "No"
```
silasary: I feel like the default here should be Yes.

nicopop: sure, thats one of the options I wasnt sure

src/ManualClient.py
```diff
@@ -155,19 +156,31 @@ class ManualContext(SuperContext):
...
         'category_odd_default': [1.0, 1.0, 1.0, 0.0],
         'category_in_logic': [2/255, 82/255, 2/255, 1],
+        'category_in_glitched_logic': [172/255, 179/255, 81/255, 1],
         'deathlink_received': [1, 0, 0, 1],
         'deathlink_primed': [1, 1, 1, 1],
         'deathlink_sent': [0, 1, 0, 1],
         'game_select_button': [200/255, 200/255, 200/255, 1],
         'header_background': [15/255, 80/255, 112/255, 1]
     }
+    def get_colors_hex(self, color: str) -> str:
```
nicopop, line 176: I plan to remove this function since it's unused but otherwise the rest is done

nicopop (commented):
I only have this to change and This PR will be fully ready for reviews

PR #210, opened 2026-03-25
Add starting items to ManualWorld to allow interacting with in hooks

fraxker:
This PR adds the ability to set starting items from within hooks. Useful when there are items that the player could start with or could be placed into generation depending on settings.

```py
def before_create_items_starting(item_pool: list, world: World, multiworld: MultiWorld, player: int) -> list:

    ### Handle Modifications to the Starting Inventory
    # for x_location, a value of 0 means starting inventory, hence the "not"
    option_item_pairs = [
        (not(get_option_value(multiworld, player, "mark_location")),"Teleport-Mark"),
        (not(get_option_value(multiworld, player, "mount_location")),"Slot-Mount")
    ]

    for option, item in option_item_pairs:
        starting_item_block = format_starting_item_block(item)
        if option: # add to starting item if option enabled
            world.starting_items.append(starting_item_block)
        elif starting_item_block in world.starting_items: # remove if option not enabled (prevents issues with multiple worlds overlapping)
            world.starting_items.remove(starting_item_block)
```

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -66,6 +66,7 @@ class ManualWorld(World):
     location_name_to_location = location_name_to_location
     location_name_groups = location_name_groups
     victory_names = victory_names
+    starting_items = starting_items
 
     event_name_to_event = event_name_to_event
 


silasary (commented):
I think you're missing some commits?

silasary:
I know how scoping works.  But starting_items is a dict, which is a mutable type. 

You're not actually changing anything by changing the scope of a reference variable.  It's still the same globally defined dict, and modifying it when you have multiple slots with different yaml settings will not do what you want it to do.

At the bare minimum, you need to be using [`copy.deepcopy`](https://docs.python.org/3/library/copy.html) to make starting_items safe to modify.

But I'd much rather have it as either a new hook, or a parameter on `before_create_items_starting`, rather than adding an undocumented field on the world object.

FuzzyGamesOn:
Re-phrasing Sila's feedback, which I agree with:

> I assumed this was a straight forward change that honestly seemed like an oversight as it is the only one of the three imports from game.py that is not saved on the world object (but is referenced inside it).

Game name and filler item name don't change on a per-slot basis. Or, if dynamic filler item names are preferred, there's a hook to handle that.

So you'd need a similar approach for starting_items, since having those change in different ways for multiple slots in the same multiworld won't work. Think of it as being more similar to item_table, etc., which come in pre-generation the same but are modified via hooks.

The right way is probably a new hook similar to `hook_get_filler_item_name`, so that the existing signature of the `before_create_items_starting` hook wouldn't need to change and break backwards compat.

PR #214, opened 2026-04-07
Client: Category sorting & Goal protection

nicopop:
like my #201 PR this pr make some misc changes but to the client this time
things included in this PR so far:
- Add the missing Click protection to the victory button
- Add Settings to sort the categories via Natural sort and/or reversed order if thats what the player wants
  - Done via 2 new settings in the F1 menu (see second screenshot)
  - Default is set to Natural
- Made location buttons check their state via a new button.location_name for UT checks and location sending
  - This is a preparation for a future PR where we can add support for location aliases either by supporting UT own alias system or from our own system (or a mix of both)
- Similarly to locations I made items use a custom label that saves their name inside of a property instead of their text
- Searches still use the button text so filtering will work with the potential future extra text
image

<details><summary>Alias stuff was moved to other branch on silasary's request but the old changelog are still here</summary>
<p>
  - I added some basic support for UT's alias system that visually added some text next to the location name (see third screenshot)
    - the support is Hook only currently but a bigger PR could probably make this better
  - This let met add an equivalent to the location's alias system (which is not a thing in UT but it fit in our client)
The following example pictures shows my Old OuterWilds manual using those aliases to protect against spoilers
image
</p>
</details>

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -179,7 +179,7 @@ class ManualContext(SuperContext):
 
         self.send_index: int = 0
         self.syncing = False
-        self.game = game
+        self.game: str = game
         self.username = player_name
 
     async def server_auth(self, password_requested: bool = False):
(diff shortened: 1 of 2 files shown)


silasary:
It's a bit hard to review with assorted misc changes all bundled together.  

Can I ask you to split the Aliases stuff into its own PR, so this one is a bit more manageable?

nicopop:
@silasary 
ill try to remove the Alias stuff from this PR Idk how to do it cleanly so ill just push a version with all its code removed

> It's a bit hard to review with assorted misc changes all bundled together.
> 
> Can I ask you to split the Aliases stuff into its own PR, so this one is a bit more manageable?

silasary (approved):

FuzzyGamesOn (approved):

PR #215, opened 2026-04-09
Fix for place_item and local item not being done at the right time during gen

nicopop:
This PR fixes multiple things that AP's Unit tests says we are to late in generation
- placement of locked items should be done at the latest in `create_items`
- item being local (or not) should be decided at the latest in `generate_early`

Moving the item placement made the after_generate_basic hook useless so I removed it
and added a before_create_items_place_items instead

To make sure we catch everything like this in the future I added test location and items that each test a property we hadn't tested before


silasary (dismissed):

FuzzyGamesOn (dismissed):
Breaking change for a couple reasons, so we'll want to let people know when it gets released.

FuzzyGamesOn (dismissed):

silasary:
Merge Conflict

silasary (approved):

FuzzyGamesOn (approved):

PR #218, opened 2026-05-12
Allow people to use the name of an existing item for their automatic filler.

silasary:
People have long desired the ability to add a category to the automatic filler.  So far, that's only been possible with hooks.

This refactors the code a bit, so that we can take the name of an existing item in the `filler\_item\_name`, which will allow for better customization.

### DIFF
--- a/src/Items.py
+++ b/src/Items.py
@@ -15,11 +15,7 @@ lastItemId = -1
 
 count = starting_index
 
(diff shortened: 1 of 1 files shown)


nicopop (approved):
looks good.

FuzzyGamesOn (approved):

PR #219, opened 2026-05-30
Item/location description in the client

nicopop:
## Added for both items.json and locations.json:
- A new optionnal string property: `description`
  - anything written in it will be displayed on the client next to the object name (see pictures below)
  - its done in such a way that location description also get added in the Tracker tab (location_id_to_alias in init.py)
  - A more dynamic description can also be added by hooks via slotdata for both location_name_to_description and item_name_to_description

Someone that knows more about kivy could potentially move the description below the item/location name but thats not something I know how to do other than maybe adding a \n

The following example pictures shows my Old OuterWilds manual using those aliases to protect against spoilers
image

<details><summary>old header</summary>
<p>
Used to be part of #214 now standalone and with a real implementation with it
it will require #214 to be merged before this one since this branch is based on it
</p>
</details>

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -148,6 +148,8 @@ class ManualContext(SuperContext):
     deathlink_out = False
 
     visible_events: dict[str, dict[str, Any]]  = {}
+    location_id_to_alias: dict[str, str] = {}
+    item_id_to_alias: dict[str, str] = {}
 
     search_term = ""
     items_sorting = SortingOrderItem.default.name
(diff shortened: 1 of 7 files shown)


silasary:
These fields should populate `webworld.location_descriptions` and `webworld.item_descriptions` respectively.

nicopop:
> These fields should populate `webworld.location_descriptions` and `webworld.item_descriptions` respectively.

@silasary 
I didnt know those were a thing... I guess I can either make the switch from world... to webworld or do like the UT alias support and just add those pointers

nicopop:
webworld.location/item_descriptions are dict[str (name), str(desc)]
while UT (and my current implementation) is dict[int (id), str (desc)] 
I was using id so it would be less data used to store thing and less text that get displayed in client for the slotdata version but I guess I can convert it all to [name, desc] and add a converstion to [id, desc] for UT

nicopop:
> Someone that knows more about kivy could potentially move the description below the item/location name but thats not something I know how to do other than maybe adding a \n

I tried just adding a \n and locations ends up looking like the following image even when I make it change height so ill leave this to a better kivy dev
(items look fine though somehow)
image

```python
                        extra = f'\n({description})' if (description := self.ctx.get_location_description_by_name(location_name)) is not None else ''
                        text = f"{location_name}{extra}"
                        location_button = TreeViewButton(text=text,
                                                         halign="center",
                                                         size_hint=(.75 if has_hint else 1, None),
                                                         height=dp(200) if extra else dp(30))
 ```
this height param doesnt seem to change much

**Comment by FuzzyGamesOn** [comment]
Seems maybe related to this feature request: https://discord.com/channels/1097532591650910289/1220087744076845226

File: src/ManualClient.py [hunk 1]
```diff
@@ -219,6 +221,20 @@ def suggested_game(self) -> str:
         from .Game import game_name  # This will at least give us the name of a manual they've installed
         return Utils.persistent_load().get("client", {}).get("last_manual_game", None) or game_name
 
+    def get_location_description_by_name(self, name: str) -> str|None:
+        # First we try to get it from slotdata for dynamic descriptions
+        description = self.location_name_to_description.get(name, None)
+        # Secondly we try to get it from the world itself for a more static description
+        if description is None:
+            description = AutoWorldRegister.world_types[self.game].web.location_descriptions.get(name, None)
```
nicopop, line 229: Just thought that I should verify if web.location_description exists by default (so it doesn't error on older manuals)
Ill come back and remove this comment if it does

nicopop: default web create empty dicts there so that should be fine

PR #220, opened 2026-06-14
Graphical tracker

lafitteque:
This PR adds the possiblility to put two new jsons in the data folder : 

- tab_items.json 
- tab_locations.json

Each json adds a corresponding tab showing a visual tracker for tab_items and a visual location sender for tab_locations.

The json format accepted is the same as the following examples :

**1. tab_items :** 
```
{
    "name": "E2", // optional
    "image_locked": "data/images/locked.png", // optional
    "image_unlocked": "data/images/base.png", // optional
    "x" : 103,
    "y" : 301,
    "width": 64,
    "height": 64,
    "label" : "E2" // optional
}
```

or 

```
{
    "image": "data/images/empty.png", // optional
    "x" : 103,
    "y" : 301,
    "width": 64,
    "height": 64,
    "label" : "E2" // optional
}
```

Items are tracked by their name. So if there is no name, it serves only as a visual help (text / image)

**2. tab_locations  : (pretty similar)**

```
{
    "name": "D6_hard", // optional
    "image_locked": "data/images/not_sent.png", // optional
    "image_unlocked": "data/images/sent.png", // optional
    "x" : 1166,
    "y" : 228,
    "width": 64,
    "height": 64,
    "label_if" : { // optional
        "easy_mode": "6s", // optional
        "normal_mode" : "8s",  // optional
        "hard_mode": "10s" // optional
    }
}
```

The tabe will create buttons based on these descriptions. If name is not given, it will only add label and image (visual help).

**Result :** When the corresponding json files are added, two new tabs appear. One showing the items received, one showing the buttons to send locations. 

**Why this should be added to the project : ** The current button lists are not immersive and can be hard to navigate or even to design when referring to things with non obvious names. In addition, graphical tabs would be a perfect way to help manual makers design their client as they will.

**Progress to be done  :** 

- A better management of window size
- More flexible json
- Add logic possibilities in the jsons
- optimizations I imagine
- Catching exceptions in the code for error feedback

EDIT : 
I forgot to mention there is no example for the the current example.
I can send you my current example I'm working with.

Also, I made no documentation for making the jsons.

EDIT 2 : 
Data folder to test the PR : [PRdata.zip](https://github.com/user-attachments/files/29037119/PRdata.zip)

### DIFF
--- a/src/Data.py
+++ b/src/Data.py
@@ -37,7 +37,8 @@ class ManualFile:
 
         return contents
 
-
+tab_items_table: list[dict[str, Any]] = convert_to_list(ManualFile('tab_items.json', list).load(), 'data') #list
+tab_locations_table: list[dict[str, Any]] = convert_to_list(ManualFile('tab_locations.json', list).load(), 'data') #list
 game_table: dict[str, Any] = ManualFile('game.json', dict).load() #dict
 item_table: list[dict[str, Any]] = convert_to_list(ManualFile('items.json', list).load(), 'data') #list
 location_table: list[dict[str, Any]] = convert_to_list(ManualFile('locations.json', list).load(), 'data') #list
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -250,12 +250,11 @@ class ManualContext(SuperContext):
     def on_package(self, cmd: str, args: dict):
         super().on_package(cmd, args)
 
-        self.ui.current_package = args
-
         if cmd in {"Connected", "DataPackage"}:
             if cmd == "Connected":
                 Utils.persistent_store("client", "last_manual_game", self.game)
                 if args.get("slot_data"):
+                    self.ui.user_options = args.get("slot_data")
                     goal = args["slot_data"].get("goal")
                     if goal and goal < len(self.victory_names):
                         self.goal_location = self.get_location_by_name(self.victory_names[goal])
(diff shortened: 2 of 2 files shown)


silasary:
Please share a working example.  It's very hard to test the code without one.

lafitteque:
Of course, here is my own example (the logic is not complete but it should not matter for testing). I just put the data folder as it's the only addition needed to test the PR.
[data.zip](https://github.com/user-attachments/files/29037019/data.zip)

FuzzyGamesOn:
This feels like it's reinventing the PopTracker wheel to some extent, since PopTracker also works with Manual sends. Did you try PopTracker for your use case, and what makes this approach preferable?

lafitteque:
Hello, I did not know about PopTracker and it definitely looks like the option to add an "item" tab is exactly the same as using PopTracker. 

Based on what I could find about PopTracker, it does not look like you can send a location from the visual guide (map or other). The "locations" tab this PR adds makes this possible.

Let me know if I am wrong. If PopTracker lets players track *and send items* then using PopTracker seems like a great solution for what I am trying to solve.

FuzzyGamesOn:
Yep, first bullet under "App Features" in this changelog entry :smiley: https://github.com/black-sliver/PopTracker/blob/master/CHANGELOG.md#v0270

lafitteque:
I have to try it, then. I'll see if it fulfills the needs I had.

Thanks for letting me know. I think it would be great to make this solution appear somewhere on the documentation.

Edit : if it's already written somewhere, my bad.

FuzzyGamesOn:
We added docs about PopTracker-related things to hopefully help with this in the future.

Last comment was a month ago. Assuming you found a solution in PopTracker, so closing. If you determine that there's a continued need for this, feel free to re-open.

PR #223, opened 2026-07-07
Add support for assigning a list of items to filler_item_name

Seatori:
Changes `filler_item_name` (and `hook_get_filler_item_name`) to allow them to be provided a list of strings instead of the standard single string, in addition to maintaining support for using a single string.

When given a list of strings, `get_filler_item_name` picks an item at random each time it is prompted to make a filler item.

I tested this by generating worlds with unique single filler names, unique lists of filler names, and filler names overlapping with items that had been assigned in `items.json` in every combination I could think of.

### DIFF
--- a/src/Items.py
+++ b/src/Items.py
@@ -30,8 +30,12 @@ for key, val in enumerate(item_table):
     item_table[key]["progression"] = val["progression"] if "progression" in val else False
     if isinstance(val.get("category", []), str):
         item_table[key]["category"] = [val["category"]]
-    if item_table[key].get("name") in filler_item_name:
-        filler_found.add(item_table[key].get("name"))
+    if isinstance(filler_item_name, list):
+        if item_table[key].get("name") in filler_item_name:
+            filler_found.add(item_table[key].get("name"))
+    elif isinstance(filler_item_name, str):
+        if item_table[key].get("name") == filler_item_name:
+            filler_found.add(item_table[key].get("name"))
 
     count += 1
 
--- a/src/Items.py
+++ b/src/Items.py
@@ -15,7 +15,7 @@ lastItemId = -1
 
 count = starting_index
 
-filler_found = []
+filler_found: set[str] = set()
 
 # add sequential generated ids to the lists
 for key, val in enumerate(item_table):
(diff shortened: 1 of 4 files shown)


src/__init__.py
```diff
@@ -75,7 +75,16 @@ class ManualWorld(World):
...
+        if isinstance(hook_get_filler_item_name(self, self.multiworld, self.player), str):
+            return hook_get_filler_item_name(self, self.multiworld, self.player)
+        elif isinstance(hook_get_filler_item_name(self, self.multiworld, self.player), list):
+            hook_random_filler_name = self.random.choice(hook_get_filler_item_name(self, self.multiworld, self.player))
+            return hook_random_filler_name
+        elif isinstance(filler_item_name, str):
+            return self.filler_item_name
+        else:
+            random_filler_name = self.random.choice(filler_item_name)
+            return random_filler_name
```
silasary: You're calling the hook twice, and discarding the first result.  If a world is counting how many times it's provided a given filler item (One of the main reasons to use this hook), counts will be wrong.


```suggestion
        hook_result = hook_get_filler_item_name(self, self.multiworld, self.player)
        if hook_result and isinstance(hook_result, str):
            return hook_result
        elif hook_result and isinstance(hook_result, list):
            return self.random.choice(hook_result)
        elif isinstance(filler_item_name, str):
            return self.filler_item_name
        else:
            random_filler_name = self.random.choice(filler_item_name)
            return random_filler_name
```

src/__init__.py
```diff
@@ -126,7 +135,7 @@ def create_items(self):
         items_config: dict[str, int|dict[ItemClassification | str | int, int]] = {}
         for name in configured_item_names.values():
             if name == "__Victory__": continue
-            if name == filler_item_name: continue # intentionally using the Game.py filler_item_name here because it's a non-Items item
+            if name in filler_item_name: continue # intentionally using the Game.py filler_item_name here because it's a non-Items item
```
silasary: This needs to be type checked.  When used with `str in str`, it's a substring search.   
As an example, a filler of "1 Coin" would accidentally delete an item called "1001 Coins".

src/Items.py
```diff
@@ -15,7 +15,7 @@
 
 count = starting_index
 
-filler_found = False
+filler_found = []
```
silasary: Minor nitpick, but a set would be more optimized than a list.

silasary (changes requested):
Mostly good, just a few minor things

silasary (approved):

silasary (approved):

FuzzyGamesOn (approved):

PR #229, opened 2026-08-25
Client: Category sortkey

nicopop:
this PR add the sort-key to categories
It also remove the "custom" sorting type we had and replaced it boolean settings
that removes the confusion of a having an alphabetical sorting that doesnt use sort key while having the natural sort that uses sort key.

### DIFF
--- a/src/ManualClient.py
+++ b/src/ManualClient.py
@@ -36,8 +36,8 @@ if typing.TYPE_CHECKING:
     import kvui
 
 class SortingOrderLoc(IntEnum):
-    custom = 1
-    inverted_custom = -1
+    # custom = 1
+    # inverted_custom = -1
     alphabetical = 2
     inverted_alphabetical = -2
     natural = 3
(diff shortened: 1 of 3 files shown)


nicopop:
ill commit a removal of the commented out removed code later

FuzzyGamesOn (approved):

src/ManualClient.py
```diff
@@ -503,6 +536,14 @@ def build_settings(self, settings: Settings):
                             "options": list(SortingOrderItem._member_names_),
                             "desc": '\n'.join([f'[b]{i.name}/inverted_{i.name}[/b]: {i.__doc__}' for i in SortingOrderItem if i.__doc__ is not None])
                         },
+                        {
+                            "type": "bool",
+                            "title": "Items Sorting uses sortkeys",
+                            "desc": "Should Item sorting respect sort-keys",
```
silasary, line 542: I'm not a fan of putting sort-keys in user-facing text.  But I can't think of better wording right now, and will PR something better when I think of it.

silasary (approved):

PR #231, opened 2026-08-28
pick_and_choose documentation can fail to load a manual if directly copied

willekeh:
in the documentation, item_name_groups was imported from ..Items  in the wrong location. because of this a manual can fail to load, if following the instructions


silasary (approved):
Thanks!

nicopop (approved):

PR #237, opened 2026-09-11
add a doc string to ManualDeathLink

nicopop:
According to Archipelago's Unit tests Deathlink should always have a doc string
In some situation the base doc string doesnt load properly and fail tests,
just adding the default string here fixes it

### DIFF
--- a/src/Options.py
+++ b/src/Options.py
@@ -22,7 +22,7 @@ class GenerateRegionDiagram(Toggle):
     visibility = Visibility.none  # Hidden option
 
 class ManualDeathLink(DeathLink):
-    pass
+    """When you die, everyone who enabled death link dies. Of course, the reverse is true too."""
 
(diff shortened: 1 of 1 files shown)


nicopop:
can't seem to reproduce issue outside of a custom manual so disregard for  now

silasary:
Tests only fail if Deathlink is enabled in game.json, but I did reproduce it.

silasary (approved):

nicopop:
sorry for this, yesterday the moment I made this PR I got hit with major (impostor syndrome level of ) doubt and then spent 40 minute trying to reproduce the issue with the latest src and couldnt get the bug to show, its nice that you also got it to fail and that it probably wasnt only something that happened in my Terraria TBG manual

FuzzyGamesOn (approved):
Good call, much better than a pass.

PR #241, opened 2026-09-14
Don't run is_location_enabled when UT is doing a fake gen

silasary:
One of the most common reasons UT fails is people using RNG to remove locations.  The recommended fix is for the user to put a fake gen check in their hooks.

This PR gets ahead of the problem and gives every location to UT regardless, and lets UT handle the checks that are missing serverside.

I also finally got around to fixing the `after_create_regions` hook to stop recommending that people remove locations after they've been created.  

This has been obsolete for a long time, and people trying to do it there has been a frequent source of IterationErrors in #manual-dev.


FuzzyGamesOn (approved):
Yep. Will be good to see some reduction in reckless `next()` calls.

nicopop (approved):

PR #245, opened 2026-09-18
Simple script to update the Client Version whenever we update the client

silasary:
See #243


FuzzyGamesOn:
Love the automation idea, not a fan of an action that modifies Python source, even if it is just a date. The AST code that finds it is fun though, lol

We should probably pull out client version into a manifest file first.

silasary:
It is done through a PR, which will still need two people to sign off on.

I'm confident enough with it, and it's never going to do anything destructive without oversight. 

Moving it to the manifest is better, but we still don't have a proper manifest implementation, and I'd rather a good solution now than wait for a perfect solution later.

FuzzyGamesOn:
We disagree on this being a good solution. :smile: 

I'll see if I can get a manifest PR together tonight or this weekend. (I should clarify: I'm not talking about an archipelago.json manifest. Just a project manifest we can use for the overall project.)

silasary:
I'd rather the effort be spent on #196 and making this a non-issue entirely.

I've changed this PR to write to a one-line file, and just importing the version from that.  It should alleviate your concerns.

